PR #1, opened 2017-03-31
Initial pact-python implementation

matthewbalvanz-wf:
- Supports creating consumer driven contracts using an external Pact mock service
- Includes the EachLike, SomethingLike, and Term matchers
- Is setup to package itself as `pact-python`

This is mostly a direct copy of the project we created internally at Workiva. Because of that its larger than I prefer and we can break things up if that is desired.

There are a couple of things worth discussion:
- The project is currently setup to enforce the code linting standards for the team that created it, which is 100% test coverage, linting of code to the [PEP-8](https://www.python.org/dev/peps/pep-0008/) and docstrings to the [PEP-257](https://www.python.org/dev/peps/pep-0257/) standard. I'm open to discussion here, but we've found it leads us to write consistently organized and reasonably well documented code.
- I added my name and email address as author in the `setup.py` defining the project. While I did write the majority of the code, it feels like it might be appropriate to have something like `Pact Foundation` and an email of some sort, but I'm unsure what.


mefellows:
To my untrained eye, this looks to be a pretty good start. Even if not perfect, I'd be inclined to pull this straight in to master as it gives us something to work from.

@jslvtr - what do you think?

pact/matchers.py
```diff
@@ -0,0 +1,196 @@
...
+    :param term: The term to be parsed.
+    :type term: None, list, dict, int, float, str, unicode, Matcher
+    :return: The JSON representation for this term.
+    :rtype: dict, list, str
+    """
+    if term is None:
+        return term
+    elif isinstance(term, (six.string_types, int, float)):
+        return term
+    if isinstance(term, dict):
```
jslvtr: Should this be `elif`?

pact/pact.py
```diff
@@ -0,0 +1,244 @@
...
+        :type query: dict, basestring, or None
+        :rtype: Pact
+        """
+        self._request = Request(
+            method, path, body=body, headers=headers, query=query).json()
+
+        return self
+
+    def will_respond_with(
+            self, status, headers=None, body=None):
```
jslvtr: Is this newline here part of the style guide? Seems that full-length 106 would be shorter than e.g. 81.

jslvtr:
Looks very good! I've put a couple of single comments in there just to satisfy my curiosity; other than that looks like the best start. 🚢

pact/matchers.py
matthewbalvanz-wf: Nice catch. This one made it through all our internal reviews. With the return statements its functionally the same, but for consistency I think its best to change it to `elif`.

pact/pact.py
matthewbalvanz-wf: My hunch is that during development I had more arguments before I fully understood the mock service. I'll shorten this up to one line.

PR #2, opened 2017-04-07
Package the Ruby Mock Service and Verifier

matthewbalvanz-wf:
- Build process now downloads and prepares the Ruby applications like the NPM based Pact
- Resulting *.tar.gz now includes portable Ruby versions of Mock Service and Verifier
- The platform appropriate version is unpacked when a user installs pact-python

@mefellows @jslvtr 

This work is based on the work in [pact-provider-verifier-npm](https://github.com/pact-foundation/pact-provider-verifier-npm/tree/master/scripts) and [pact-mock-service-npm](https://github.com/pact-foundation/pact-mock-service-npm/tree/master/scripts) and packages the same 4 platforms of each as part of the Python tar file. This results in a fairly big file (nearly 70MB), but the only way I've spotted without requiring the user to have Ruby is to combine the mock-service and verifier into a single install so we can cut the number of traveling Ruby installs in half. I don't think I have enough Ruby skills to know if that's possible and how it might be done yet. If you guys have ideas let me know, otherwise we might leave that as an optimization exercise in the future, but the Rust version might beat us there and we can switch to that instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # pact-python specific ignores
 e2e/pacts
+pact/bin
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,7 @@ sudo: false
 
 install: pip install -r requirements_dev.txt
 
-script: make test
+script:
+  - flake8
+  - pydocstyle pact
+  - tox --develop
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,6 @@
 include LICENSE
+include *.gz
 include *.txt
 include *.md
 prune *test
+prune e2e/*
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ endef
 
 help:
 	@echo ""
+	@echo "  clean      to clear build and distribution directories"
 	@echo "  deps       to install the required files for development"
 	@echo "  package    to create a distribution package in /dist/"
 	@echo "  release    to perform a release build, including deps, test, and package targets"
@@ -19,6 +20,13 @@ help:
 
 .PHONY: release
 release: deps test package
+
+
+.PHONY: clean
+clean:
+	rm -rf build
+	rm -rf dist
+	rm -rf pact/bin
 
 
 .PHONY: deps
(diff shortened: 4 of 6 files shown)


matthewbalvanz-wf:
I've rebased this PR on master so the Travis CI configuration is included. At the moment I have it setup to only run tests and not to build the Python package as the version of Ruby installed in the Python VM is only 1.9.

In the future, we may want to investigate the options available from Travis to see if we could create a single artifact that is the tar.gz file from the build that includes the Ruby mock-service and verifier applications.

jslvtr:
This looks awesome. I'm fairly confident Travis should be able to do what you suggest!
🚢

mefellows:
Sorry for my tardiness, i'll have some time to look through this early next week in detail I'd say. Don't wait for my nod of approval or anything, it's trending in the right direction. I can probably lend a hand to the Ruby stuff.

I don't know enough about the Pip packaging ecosystem to guide you here, I'll do some research before I make concrete suggestions. The idea of packaging both Ruby processes makes sense though. In the node versions, we package all 4 platforms but publish 4 separate (actually 5) packages, one as a meta-package and 4 platform specific packages. 

This way, users don't pay the cost of having a 'universal' package, if that makes sense.

FWIW your bash fu is much nicer than the existing, well done.

matthewbalvanz-wf:
Your point about the node packaging got me thinking again. I might be able to assemble something similar in the Python world. Originally, I had discounted that idea as I thought Travis would prevent us from publishing multiple packages from the same repository, but perhaps that is just me limiting my thinking based on my usual tools as the node version accomplishes the same in its [`.travis.yml`](https://github.com/pact-foundation/pact-mock-service-npm/blob/master/.travis.yml).

I think that would be a great improvement for users and is something I would want to complete for the 1.0.0 release. I added a card to the [project board](https://github.com/pact-foundation/pact-python/projects/1) where I've been collecting my thoughts on the subject. I'm open to rolling it into this work or spinning it off to another PR though.

mefellows:
Great!

I personally am happy with smaller PRs that each contribute working software, so no objections for me to roll this in first and then improve later.

jslvtr:
Definitely agree, and this will make it easier to test locally for development.

PR #3, opened 2017-04-07
Add Travis CI build

matthewbalvanz-wf:
- Setup a Travis CI configuration to perform a "build" of pact-python
- Run the tests using tox in Python versions 2.7, 3.3, 3.4, 3.5, and 3.6.

@jslvtr I threw this together so I don't accidentally commit broken things. We should be able to expand it into a full build process as we go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+python:
+  - "2.7"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+  - "3.6"
+
+sudo: false
+
+install: pip install -r requirements_dev.txt
+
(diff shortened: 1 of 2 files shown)


jslvtr:
Great stuff! Thanks @matthewbalvanz-wf for adding this. We're aiming for 100% coverage throughout, right?

matthewbalvanz-wf:
@jslvtr That's my goal, and I think coverage is checking that at the moment, I think that's buried in the build logs though.

jslvtr:
Sounds like a plan! We can check manually for now and surface it after. I'll look into it.

Issue #7, opened 2017-04-17
Allow pacts to be stored in a specified directory


kpodl:
Hi there,
First of all: Great, that you have created `pact-python`! I was just in the process of trying to resuscitate another pact implementation for Python. Yours, however, is much nicer.
Here is my feature request in story form (as you seem to prefer them that way):

As a user creating consumer contracts,
I want to be able to specify a directory in which pacts, that have been successfully verified are stored,
and I want that pact files are named "sensibly" and that name clashes are automatically resolved,
So that I do not have to use a docker volume to collect pacts.

The idea would be to allow a directory to be specified in the `has_pact_with()` method which will be passed to the `Pact` instance. Then the pact returned by the `POST` to `/pact` in `Pact.__exit__()` can be saved to a file.
If there are no objections, I would be happy to submit a pull request!

jslvtr:
Thanks for this! May be interesting to think of them as two separate features—not that it matters much if your PR will solve for both.

> As a user creating consumer contracts, I want pact files to be named sensibly and without clashing with existing files, so that there are no errors, missing files, or overwritten files.

> As a user creating consumer contracts, I want to be able to specify which folder to save successfully verified pacts, so that I don't have to use a docker volume to collect pacts

What do you think? I'm not too experienced with this, so not sure how this would play with pact brokers, maybe @mefellows can shed some light on it

mefellows:
The naming of the pact files is standardised and based entirely on the name of the consumer and provider. The underlying Ruby Mock Service will automatically generate the correct file name, I don't believe you can control this from Python (happy to be corrected though!).

There _is_ an edge case where if two services are named similarly, the Broker will complain - but this is very much a separate use case, and all that can really be done from the client (Python lib) is to propagate the error back to the user in a nice way.

As for the placement of the Pact files, yes, the output directly should be configurable (the Mock Service takes an argument `--pact-dir` that will save them where specified)

matthewbalvanz-wf:
Hi @kpodl! I agree that the user experience could use some tweaks. Based on my work over in #2 I had started to move away from Docker and get the various options from the Ruby mock service implemented in the Python API, which includes `--pact-dir`. I'm hoping to get some time this week to finish it up, but my day job keeps getting in the way ;-)

My current work is over in [this branch](https://github.com/pact-foundation/pact-python/compare/manage-mock-service?expand=1) if you are interested in taking a look.

matthewbalvanz-wf:
[...]

PR #8, opened 2017-05-02
pact-python manages the mock service

### LINKED ISSUE #5: pact-python starts and stops the mock service
As a user creating consumer contracts using pact-python,
I want pact-python to manage the mock service for me,
so that using pact-python is greatly simplified for me.

matthewbalvanz-wf:
- pact-python now manages the mock service process
- Moved the project to version v0.2.0

resolves #5 and possibly #7 

@mefellows @jslvtr

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,1 +1,2 @@
 v0.1.0, 2017-03-30 -- Basic support for authoring contracts against a separately running mock service
+v0.2.0, 2017-04-22 -- Pact Python manages the Pact mock service
--- a/Makefile
+++ b/Makefile
@@ -39,9 +39,6 @@ e2e:
 	sh -c '\
 		cd e2e; \
 		docker-compose pull > /dev/null; \
-		docker-compose up -d pactmockservice; \
-		while ! nc -z localhost 1234; do sleep 0.1; done; \
-		docker-compose logs --follow > ./pacts/mock-service-logs.txt & \
 		nosetests ./contracts; \
 		docker-compose down; \
 		docker-compose up -d app pactverifier; \
--- a/README.md
+++ b/README.md
@@ -40,14 +40,19 @@ def user(user_name):
 Then `Consumer`'s contract test might look something like this:
 
 ```python
+import atexit
 import unittest
 
 from pact import Consumer, Provider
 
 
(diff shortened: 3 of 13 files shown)


mefellows:
You're a _machine_! Great stuff.

mefellows:
One thing to consider is when the verification call is made. You may want to consider exposing this to the DSL somehow to the user. In larger Pact tests, if you have several request/response related unit tests but one of them doesn't call the mock service correctly, you won't know if it fails until the process exists (based on what I can glean, this occurs using`atexit` and https://github.com/pact-foundation/pact-python/blob/4bf7b8b199736df2c40420908f5b3843acae712d/pact/pact.py#L221).

This could make it hard to see why or where the test failed (i.e. which tests didn't send through the Interaction or sent an unexpected one).

matthewbalvanz-wf:
> One thing to consider is when the verification call is made...

Python has a concept called [context managers](http://book.pythontips.com/en/latest/context_managers.html) that allow you to run code before and after a block, for example:

```python
pact = Consumer('Consumer').has_pact_with(Provider('Provider'))
with.pact:
    # perform my tests
```
Before my tests `__enter__` is called, after `__exit__` is called. `__exit__` performs a verification. You do make me think that might not be flexible enough, and that using `start`/`stop` as the methods to launch and teardown the Ruby mock service may lead to confusion.

I'll ponder that for a bit and may rearrange some things to make it more clear.

mefellows:
Is that at the test class level or individual tests? Another thing to consider is that people may break up tests to a single provider across files as pact tests can be quite lengthy, and that they'll need to use the same running mock server for all of them (this will be made easier when we get the shared lib). That being said, having an opinionated workflow isn't necessarily a bad thing.

matthewbalvanz-wf:
@mefellows I pulled out the setup/verify logic to separate methods so users now have two ways to use the `Pact`:
1. As a context manager with the `with` statement, which calls `setup` before running the code in the block, and `verify` after the block completes
2. Manually calling `setup` and `verify` in their code.

I also renamed the `start` and `stop` methods to `start_service` and `stop_service` to make it more obvious what they were doing.

mefellows:
That seems like a great compromise @matthewbalvanz-wf!

Issue #10, opened 2017-05-12
Upgrade Pact Mock Service


matthewbalvanz-wf:
The version of Mock Service used during the build of pact-python is currently out of date and does not include a recent change to make the pact specification version consistent across the various language implementations: https://github.com/pact-foundation/pact-mock_service/commit/5b89d95167013ac4b11030b362c439f5632e7540

We need to upgrade to at least 2.1.0 to ensure pact-python is creating compatible pact files.

jslvtr:
🚢 Looks good to me! Also my apologies for being unavailable, hope to have a bit more time in the coming weeks.

matthewbalvanz-wf:
No worries @jslvtr. I myself have been busy with things outside the project and haven't made as much contributions as I like.

matthewbalvanz-wf:
Resolved with #11

PR #11, opened 2017-05-13
Updated Versions of Pact Ruby applications

matthewbalvanz-wf:
- Pact Mock Service is now v2.1.0
- Pact Provider Verifier is now v1.0.1
- Fixed a disconnect in the e2e tests where the consumer wrote to a different
  than the provider tests expected

@mefellows @jslvtr FYI


mefellows:
LGTM

matthewbalvanz-wf:
Resolves #10

PR #14, opened 2017-05-20
Command line application for verifying pacts

### LINKED ISSUE #6: pact-python has a command line interface to verify Pacts
As a user with a Python based provider,
I want pact-python to offer me a way to verify pacts published by my consumers,
so that I can easily verify pacts using the same package I use to author contracts.

matthewbalvanz-wf:
- Added `pact-verifier` which is a command line application that calls the Ruby verifier
- It supports the same options as the JS version of Pact, including provider state and broker options
- Removed all Docker related requirements

@mefellows @jslvtr 

Resolves #6


pact/verify.py
```diff
@@ -0,0 +1,108 @@
...
+            + ' To use provider states you must provide both'
+              ' --provider-states-url and --provider-states-setup-url.')
+        raise click.Abort()
+
+    missing_files = [path for path in pact_urls if not path_exists(path)]
+    if missing_files:
+        click.echo(
+            error
+            + ' The following Pact files could not be found:\n'
+            + '\n'.join(missing_files))
```
kennytrytek-wf: This is unicode-safe:
```
>>> '\n'.join([u'\ufffd', 'str me'])
u'\ufffd\nstr me'
```

PR #19, opened 2017-05-29
Allow falsy responses (e.g. 0 not as a string).

ftobia:
In playing with pact-python I found I was unable to generate a pact file with a body of 0 where the 0 wasn't already a string (e.g. it's a json-encoded response of the number 0). I believe this PR rectifies the issue.

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -326,7 +326,7 @@ class Response(FromTerms):
     def json(self):
         """Convert the Response to a JSON version for the mock service."""
         response = {'status': self.status}
-        if self.body:
+        if self.body is not None:
             response['body'] = self.body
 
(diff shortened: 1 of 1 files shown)


ftobia:
Let me know if you want a regression test as part of this PR. I can try to get to it this week (I haven't stood up a pact dev environment yet, I manually verified this fix with the code I've been using to play with Pact).

Thanks for porting pact to python, so far it's really great.

matthewbalvanz-wf:
@ftobia Thanks for submitting this PR. That's definitely an issue, and I'm surprised I didn't think of it when I put it together initially, because it would fail for things like empty lists and dicts.

I can setup a test for this and get both merged into master.

PR #26, opened 2017-06-17
Fixes #25: support for stubbing multiple requests at the same time 

### LINKED ISSUE #25: API should allow multiple requests to be stubbed at the same time
Multiple interactions should be allowed to be set up at the same time. The /delete call should be called before the test, not before each interaction is set up. 

See this example using bash here: https://gist.github.com/bethesque/9d81f21d6f77650811f4

https://github.com/pact-foundation/pact-python/blob/master/pact/pact.py#L119

```python
def setup(self):
        """Configure the Mock Service to ready it for a test."""
        try:
            payload = {
                'description': self._description,
                'provider_state': self._provider_state,
                'request': self._request,
                'response': self._response
            }

            resp = requests.delete(
                self.uri + '/interactions', headers=self.HEADERS)

            assert resp.status_code == 200, resp.content
            resp = requests.post(
                self.uri + '/interactions',
                headers=self.HEADERS, json=payload)

            assert resp.status_code == 200, resp.content
        except AssertionError:
            raise
```

AnObfuscator:


pact/pact.py
```diff
@@ -103,26 +100,20 @@ def given(self, provider_state):
         :type provider_state: basestring
         :rtype: Pact
         """
-        self._provider_state = provider_state
+        self._interactions.insert(0, {'provider_state': provider_state})
```
matthewbalvanz-wf: I'm fine with the code as is, but my curiosity streak wants to ask: Was there a particular case that stood out that led to inserting the new interactions at the beginning of the list instead of using `append` and `self._interactions[-1]`?

AnObfuscator: Heh. No, there wasn't. It is just an artifact of my thought process... The list was, in my head, a stack that I was pushing/peeking.

PR #27, opened 2017-06-20
Resolves #20: Download pre-packaged mock service and verifier during installation

### LINKED ISSUE #20: Reduce the download size for users by splitting the compiled application into a separate package for each platform
The current install package for pact-python is more than 60MBs and includes the mock service and provider verifier for platforms the user is not using. We should remove the requirement that the user download those unneeded versions.

matthewbalvanz-wf:
- Switch the install scripts to download the appropriate package from the Github releases page of pact-ruby-standalone
- Add a custom `develop` command that does the same installation when installing in `developer mode`

@mefellows @jslvtr


matthewbalvanz-wf:
Resolves #20

bethesque:
Hoorah for deleted code! I'm hoping to be able to make those packages smaller too now that I've got a windows build going [here](https://github.com/pact-foundation/pact-ruby-standalone-windows-test) and can start removing files and seeing if it still works ;)

PR #30, opened 2017-06-21
Add some guidelines for raising issues

bethesque:
This should help us resolve issues more quickly.


matthewbalvanz-wf:
Thanks @bethesque!

Issue #21, opened 2017-05-31
Cannot supply multiple files to pact-verifier


kevinsimons-wf:
According to the [README](https://github.com/pact-foundation/pact-python#--pact-urls), to supply multiple local pact files you give the `--pact-urls` argument a comma separated list of files.

However, this will give you an error that the file cannot be found and that file is the combined paths, 
```
Error: The following Pact files could not be found:
/tmp/pacts/pact1.json,/tmp/pacts/pact2.json
```
when they should be separated by a [new line](https://github.com/pact-foundation/pact-python/blob/master/pact/verify.py#L69).

This is because for the click option `--pact-urls` the parameter `multiple=True` is used. However this is meant for providing multiple `--pact-urls` flags such as
```
pact-verifier --pact-urls=./provider/pact1.json --pact-urls=./provider/pact2.json
```
http://click.pocoo.org/5/options/#multiple-options

Either the documentation should be updated to list that you should provide an argument for each file, or remove `multiple=True` and just do a `.split(',')` when reading the pact-urls argument.

bethesque:
I much prefer `- -pact-url path - - pact-url otherpath`. If the underlying ruby command line tool would support that, it's what I'd have used in the ruby interface rather than comma delimited, which is ugly with URLs.

bethesque:
Sorry, ignore spaces between dashes, on phone.

matthewbalvanz-wf:
Resolved in v0.7.0

PR #32, opened 2017-06-24
Issue: Cannot supply multiple files to pact-verifier

SimKev2:
- Updated '--pact-urls' to be a single comma separated string argument
 - Added '--pact-url' which can be specified multiple times

Fixing https://github.com/pact-foundation/pact-python/issues/21

Note that this will break any existing versions that are using multiple `--pact-urls` arguments

@matthewbalvanz-wf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,9 @@ local_settings.py
 instance/
 .webassets-cache
 
+# Intellij stuff
+.idea/
+
 # Scrapy stuff:
 .scrapy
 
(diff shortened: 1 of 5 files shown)


bethesque:
When making a change to a public API, it's good practice to provide an upgrade path that does not break any existing users. You could maintain the existing pact_urls in parallel with pact_url, but print a deprecation warning to let the user know to change to the new format. The old pact_urls would then be removed in the next major release, according to semantic versioning practice.

SimKev2:
@bethesque Good idea. I re-enabled multiple for `--pact-urls` and added some comments that it can be removed in major version 1.0.0

bethesque:
👍

SimKev2:
Rebased onto master to fix the new version stuff @matthewbalvanz-wf

Issue #31, opened 2017-06-22
ImportError: No module named requests


bethesque:
Is there a missing dependency declaration, or am I just a python noob?

```
$ pip install pact-python
Collecting pact-python
  Downloading pact-python-0.6.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/c8/ypphsm1d4gq3klcygh889hjm0000gn/T/pip-build-ffGYmj/pact-python/setup.py", line 14, in <module>
        import pact
      File "pact/__init__.py", line 2, in <module>
        from .consumer import Consumer
      File "pact/consumer.py", line 2, in <module>
        from .pact import Pact
      File "pact/pact.py", line 7, in <module>
        import requests
    ImportError: No module named requests
```

hartror:
👋  @bethesque's token Python dev friend chiming in.

A side note on this issue. `tox -r` should have found this install bug but it appears that tox installs `requirements_dev.txt` which has `-e requirements.txt`. This of course means that the dependencies are installed before `setup.py` runs in tox.

matthewbalvanz-wf:
This should be resolved in v0.6.1.

PR #34, opened 2017-06-25
Changed setup.py to read version from __version__.py

hartror:
Fix for #31.

* Added test & install environments to `tox.ini` to replicate the reported issue.
* Changed the location of `__version__` and its loading using the same method requests[1] et al use.

I don't love the `eval` but this is one of the few good uses for it.


matthewbalvanz-wf:
Thanks @hartror. I concur with your feelings on `eval`, and I think this might be the first time I've found a case where I can't legitimately suggest a better alternative. I'll get this merged and released as 0.6.1 so users don't run into install issues anymore.

Issue #39, opened 2017-07-21
RuntimeError: The Pact mock service failed to start.


zsunshinexue:
** # In the _wins7 system_. After installing pact, run the error message. Here's the information. What's the problem?**
"D:\Program Files_Work\Tool_Test\python27\python.exe" D:/co-mall/python-pact-testing-example-master/src/testuser_consumer.py
D:\Program Files_Work\Tool_Test\python27\Lib\site-packages\pact\bin\mock-service\lib\ruby\bin\..\bin.real\ruby.exe: No such file or directory -- Files_Work/Tool_Test/python27/Lib/site-packages/pact/bin/mock-service/lib/app/lib (LoadError)
Traceback (most recent call last):
  File "D:/co-mall/python-pact-testing-example-master/src/testuser_consumer.py", line 9, in <module>
    pact.start_service()
  File "D:\Program Files_Work\Tool_Test\python27\lib\site-packages\pact\pact.py", line 154, in start_service
    raise RuntimeError('The Pact mock service failed to start.')
RuntimeError: The Pact mock service failed to start.

bethesque:
Can you tell me if `Files_Work/Tool_Test/python27/Lib/site-packages/pact/bin/mock-service/lib/app/lib` exists?

We have a problem starting the mock service on windows because windows does not implement fork, so even if you weren't getting this "file not found" error, it would blow up later on when it hit the fork. I'm working on a different method of starting the mock service that doesn't use fork, but at the moment, I'm having trouble killing the process gracefully.

zsunshinexue:
@bethesque  This file exists，Do you mean that Pact is not supported on wins for now?

bethesque:
You can start the mock service manually by installing the [standalone mock service](https://github.com/pact-foundation/pact-ruby-standalone/releases) which is what python uses under the hood.  You can start the mock service by running `pact-mock-service service`. This will run the process in the foreground until you quit it using ctl-c. You'll need to comment out the python code that starts and stops the mock service, and you'll need to start and stop the mock service manually each time.

If you have any expertise on sending SIGTERMs to processes on windows, please let me know [here](https://github.com/pact-foundation/pact-mock_service/issues/73)

matthewbalvanz-wf:
@zsunshinexue In pact-python 0.8.0 I reworked how the mock service is managed. My experiments running the end-to-end tests on a Windows machine were successful, and I hope that you find a similar result with the new version.

[...]

PR #41, opened 2017-07-29
Move to using the `service` command with pact-mock-service

matthewbalvanz-wf:
The start/stop commands in pact-mock-service have trouble with Windows because
it does not exactly support SIGTERM. The simplest solution is to shift how pact-python
runs the mock service to:

- Start the mock service using the `service` command
- Keep track of the subprocess during pact execution
- Use the terminate options from Python's subprocess module to stop the mock service

@mefellows @jslvtr @bethesque

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -2,9 +2,13 @@
 from __future__ import unicode_literals
 
 import os
+import platform
 from subprocess import Popen
 
+import psutil
 import requests
+from requests.adapters import HTTPAdapter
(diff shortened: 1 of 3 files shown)


matthewbalvanz-wf:
This should resolve #39

mefellows:
I'm probably going to look at this code to copy/pilfer/... to the NodeJS version. I'll be watching to see how it pans out - great stuff!

PR #46, opened 2017-08-25
make setup-url name format match above reference

bassdread:
fix-setup-url-name isn't referenced anywhere in the code, the format provider-states-setup-url is used else where in the README.md.

This PR assumes that provider-states-setup-url is the correct name and makes usage the same throughout the doc.


matthewbalvanz-wf:
Thanks for clarifying this in the README @bassdread

Issue #47, opened 2017-08-29
Requests is pinned at 2.12.3, allow it to use a later version


bassdread:
I am currently unable to build with `pact-python` as I have another library using the latest `requests` version `2.12.4` and currently `pact-python` has it pinned to `2.12.3`

Can the requirement for `requests` be changed to `requests>=2.12.3` ?

bethesque:
Is there a version specifier in python that allows patch updates like in Ruby? It would be best practice to do "~> 2.12" (allow any increments except major ie. breaking versions) in Ruby, but I'm not sure if the norms in python.

matthewbalvanz-wf:
@bassdread Sorry for the delay. I've been out of town and away from the computer.

As @bethesque points out, I should have made this (and the others) a range in the first place. I did some testing with requests and it looks like anything from 2.5.0 to the latest release should work properly. Would you be willing to update your PR to open it up with: `requests>=2.5.0,<3.0.0`?

ftobia:
I'd like to suggest a different approach to relaxing version pin in `requirements.txt`. See #60. Suggestions welcome.

richard-reece:
@matthewbalvanz-wf is there a reason to pin at all?

richard-reece:
Similarly, six is pinned to version 1.10.0 which also causes pain :-(

ftobia:
I was thinking that pact-python could vendor six (like, copy six.py into the pact-python codebase, to remove the dependency). It's a somewhat common pattern, and per [these docs](https://pypi.python.org/pypi/six) "[Six] is contained in only one Python file, so it can be easily copied into your project.", so it's at least supported.

If @matthewbalvanz-wf or someone else might be amenable to removing the six dependency and including six.py in the codebase, I'll put up a PR for it.

richard-reece:
Does it need to be, though? What benefit does vendoring it bring? What issue does doing that avoid? Why can't it just be unpinned (or at least have a far more permissive version range)?

richard-reece:
Would it be acceptable to mark the current pinned versions as minimums and then add upper limits as they arise, rather than try to predict them? The reason I say this is that a user of pact-python can always *add* a requirement to enforce their own upper limit to work around an issue, but they can't *undo* a limit that you've imposed.

matthewbalvanz-wf:
Yep, I think that's a good next step, we can move things back if we find some time to investigate further.

PR #48, opened 2017-08-29
Allow newer versions of requests

bassdread:
Fixes https://github.com/pact-foundation/pact-python/issues/47

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 click==6.7
 psutil==5.2.2
-requests==2.12.3
+requests>=2.12.3
 six==1.10.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 click==6.7
 psutil==5.2.2
-requests>=2.12.3
+requests>=2.5.0,<3.0.0
 six==1.10.0


jslvtr:
Not too sure about this—if we stick to one known version, it's less likely bugs will be introduced by future versions. I think we should only upgrade to future versions if we find we need to because the current version has a bug or new versions introduce new functionality we want.

matthewbalvanz-wf:
As a library that others are likely to include in other projects, this kind of flexibility is handy, because it offers flexibility for the project to decide which version they want to use.

I did some testing with requests and pact-python and it looks like anything from 2.5.0 to the latest release should work properly. @bassdread  would you be willing to update your PR to open it up with: requests>=2.5.0,<3.0.0? If a major version is released we'll likely want to do some checking on what changed in that release before its used with pact-python.

bassdread:
@matthewbalvanz-wf apologies for the late reply, I have updated this.

I don't mind which option is used to fixed this issue. Feel free to use this or close it if there is a better option.

matthewbalvanz-wf:
Thanks @bassdread. I'm going to merge both and spend a little time sorting out what the minimums are for the project and get a new release cut.

PR #49, opened 2017-09-04
Resolve #43: Rename SomethingLike to Like

### LINKED ISSUE #43: Suggestion: rename SomethingLike to just Like
When the SomethingLike was invented, it was a bit of a hack to help out with some brittle tests we had at REA, and I never thought it would become the standard tool it has become! In the other implementations, we're using just the name "Like" rather than "SomethingLike". It's a bit pithier. Can I suggest that you do the same for python?

matthewbalvanz-wf:
- This is step 1 of the rename, adding Like as a alias for SomethingLike
- SomethingLike will be removed in version 1.0.0

@jslvtr @mefellows @bethesque

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,24 +156,25 @@ as `generate`, in this case `2016-12-15T20:16:01`. When the contract is verified
 provider, the regex will be used to search the response from the real provider service
 and the test will be considered successful if the regex finds a match in the response.
 
-### SomethingLike(matcher)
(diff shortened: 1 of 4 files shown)


bethesque:
👍

jslvtr:
All up for this, shall we merge?

mefellows:
Yep!

PR #62, opened 2017-10-31
 Add support to publish verification result to pact broker

dhoomakethu:
Publish verification results to a pact broker


matthewbalvanz-wf:
Thanks @dhoomakethu for submitting this change. It looks like there might be a few code style issues causing the build to fail. Would you be willing to add a commit to resolve those before we merge this pull request?

dhoomakethu:
@matthewbalvanz-wf Done!

PR #67, opened 2017-12-14
Enable possibility to use Matchers in path property of Request class

jawu:
Added from_term processing of path property in the constructor of Request


matthewbalvanz-wf:
Thanks for the submission @jawu

PR #72, opened 2018-02-09
Reduce timeout in _wait_for_server_start to 25s

fabianbuechler:
Modify Retry.total in Pact._wait_for_server_start to reduce the total
wait time from 54 minutes to about 25 seconds.

With a Retry.backoff_factor of 0.1 and 15 retries, the total wait time
before an error is `sum([0.1 * (2**(i - 1)) for i in range(16)])` which
is 3276 seconds or 54 minutes.

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -256,7 +256,7 @@ class Pact(object):
         :raises RuntimeError: If there is a problem starting the mock service.
         """
         s = requests.Session()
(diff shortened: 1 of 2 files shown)


fabianbuechler:
Just added a fix for a test I broke by changing the `Retry()` call.

matthewbalvanz-wf:
Thanks @fabianbuechler !

PR #74, opened 2018-02-15
Fix #73 by adding BUNDLE_FROZEN=1 env to subprocess calls.

### LINKED ISSUE #73: MockService blocks test execution if pact-python is installed in system Python's site-packages
Package versions:
* pact-python 0.14.0
* Python 3.6.3 from official docker image

When installing pact-python into system-Python's site-packages, which is quite common when using a docker container, the Mock-Service cannot be started if the Python tests in the docker-container are not executed with root permissions.

Basically I have a test that creates a `pact.Consumer` and then calls `consumer.pact.start_service()` to run the mock service within a pytest fixture. That's roughly equivalent to:

```python
import pact

consumer = pact.Consumer('consumer').has_pact_with(
    provider=pact.Provider('provider'),
    host_name='localhost',
    mock_port=1234,
)
consumer.pact.start_service()

# Execute test here.
# Tests want to access the running mock-service.

consumer.pact.stop_service()
```

However, during `consumer.service_start()` the MockService tries to modify its `Gemfile.lock` which is located at `/usr/local/lib/python3.6/site-packages/pact/bin/pact/lib`.
When running the docker container without root permissions, this file cannot be modified.

pytest prints the error message when invoked with `-s` (no capture) flag:

```
There was an error while trying to write to Gemfile.lock. It is likely that 
you need to allow write permissions for the file at path: 
/usr/local/lib/python3.6/site-packages/pact/bin/pact/lib/vendor/Gemfile.lock
```

I posted a pull request (pact-foundation/pact-python#72) for a related problem, which is that during the call to `consumer.pact.server_start()` the `Pact._wait_for_server_start` method retries a connection to the started MockService 15 times. With the implementation of `urllib3.utils.retry.Retry` with a `backoff_factor` of `0.1` that means 54 minutes wait time before the tests actually fail.

Thus it appears as if the tests would simply block. In combination with pytest's output capturing this was pretty hard to spot. :)

I don't really know Ruby, so I cannot tell why the `Gemfile.lock` has to be modified when starting the MockService. However, when installed as a vendored library of a global Python site-package, that doesn't work.
Maybe you could download the vendored ruby libs to a temporary / cache directory?

fabianbuechler:
This keeps the `bundle exec` call from modifying the `Gemfile.lock`,
which does not work if pact-python is installed in the system Python's
site-packages.

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -158,7 +158,7 @@ class Pact(object):
         if self.sslkey:
             command.extend(['--sslkey', self.sslkey])
 
-        self._process = Popen(command)
+        self._process = Popen(command, env={'BUNDLE_FROZEN': '1'})
         self._wait_for_server_start()
 
     def stop_service(self):
--- a/pact/test/test_pact.py
+++ b/pact/test/test_pact.py
@@ -228,32 +228,40 @@ class PactStartShutdownServerTestCase(TestCase):
         with self.assertRaises(RuntimeError):
             pact.start_service()
 
-        self.mock_Popen.assert_called_once_with([
-            MOCK_SERVICE_PATH, 'service',
-            '--host=localhost',
-            '--port=1234',
-            '--log', '/logs/pact-mock-service.log',
-            '--pact-dir', '/pacts',
-            '--pact-file-write-mode', 'overwrite',
-            '--pact-specification-version=2.0.0',
-            '--consumer', 'consumer',
-            '--provider', 'provider'])
+        self.mock_Popen.assert_called_once_with(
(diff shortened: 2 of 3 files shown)


bethesque:
It's just occurred to me that I could put this in the standalone script. Let me look and get back to you.

bethesque:
OK, I've added it to the standalone. You can close this, and raise a request to update to the latest standalone. It's in 1.29.0

fabianbuechler:
Thanks, that's fine as well. I've opened issue #75 for that.

PR #79, opened 2018-03-06
Added windows specific code to check if mock service is stopped.

shahha:
I was getting error while using pact-python in windows 10 'There was an error when stopping the Pact mock service.' Same code was running without any error in Ubuntu.
While digging a bit found out that checking  self._process.returncode != 0 does not work with windows as in my case windows was returning return code as 15.
So now for windows I am checking if process still exists using psutil.pid_exists to determine if mock service is stopped or not.

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -169,13 +169,18 @@ class Pact(object):
             p = psutil.Process(self._process.pid)
             for child in p.children(recursive=True):
                 child.terminate()
+            p.wait()
+            if psutil.pid_exists(self._process.pid):
+                raise RuntimeError(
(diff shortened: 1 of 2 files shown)


bethesque:
You wouldn't believe the fun and games I've had trying to ensure the mock process shuts down nicely on Windows!

shahha:
Sorry as I am new here, do I have to inform someone to merge this PR?

bethesque:
Matthew will do it when he has time. It will be in his backlog!

matthewbalvanz-wf:
Sorry for the delay. Thanks for the contribution @shahha !

PR #87, opened 2018-06-01
Add support to custom provider header

acabelloj:
Latest version of pact standalone brings support to custom provider header, this commit upgrades the version to 1.44.3

### DIFF
--- a/pact/__version__.py
+++ b/pact/__version__.py
@@ -1,3 +1,3 @@
 """Pact version info."""
 
-__version__ = '0.16.1'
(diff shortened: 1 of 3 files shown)


matthewbalvanz-wf:
@acabelloj Thanks for PR and sorry for the delay in responding. It looks like the latest version of setuptools may have dropped support for Python 3.3 and that's causing the build to fail. If you're interested in pinning it to one that works here to get things passing that would be great, otherwise I think I can work some time into my schedule in the next couple of days to make that happen.

PR #94, opened 2018-10-20
Changed given to be optional

### LINKED ISSUE #89: `given`/provider state should not be necessary
Only `Pact.given()` inserts a new element into the `_interactions` array, meaning that if you don't call it for the *first* test, you'll get an exception when calling any other interaction-related methods, and if you don't call it for *subsequent* tests, you'll overwrite the previous interaction (possibly partially, leading to nonsense).

According to [the JSON schema](https://bitbucket.org/atlassian/pact-json-schema/src/8d0b9d2f38fe517dfbce0e5d0b8755eaefaacb5b/schemas/v2/schema.json?at=master&fileviewer=file-view-default#schema.json-16:30), `providerState` is an optional field, which makes sense, since not every interaction *will* require state to be set up.

As is, `pact-python` requires a bogus `given()` call for stateless interactions (and therefore requires the provider to handle that bogus state-setting).

(Less seriously, but still I think inelegantly, this also means that `given` has to be called *first*, even though the mock-English encouraged by the method names and chaining would suggest that there's no difference between "given …, upon receiving …, will respond with …" and "will respond with …, upon receiving …, given …"

yangineer:
Happy Hacktoberfest!
Looking to contribute to one of my favourite libraries!
This PR fixes #89 
Note: `given` needs to be come in before all three mandatory fields do.

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -38,6 +38,8 @@ class Pact(object):
 
     HEADERS = {'X-Pact-Mock-Service': 'true'}
 
+    MANDATORY_FIELDS = {'response', 'description', 'request'}
+
     def __init__(self, consumer, provider, host_name='localhost', port=1234,
                  log_dir=None, ssl=False, sslcert=None, sslkey=None,
                  cors=False, pact_dir=None, version='2.0.0',
@@ -114,7 +116,8 @@ class Pact(object):
         :type provider_state: basestring
         :rtype: Pact
         """
-        self._interactions.insert(0, {'provider_state': provider_state})
+        self._insert_interaction_if_complete()
+        self._interactions[0]['provider_state'] = provider_state
         return self
 
     def setup(self):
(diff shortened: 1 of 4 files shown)


yangineer:
Hmm... looks like the latest version of Click broke the test.
I'll make a separate MR for it.

lextiz:
@matthewbalvanz-wf @bethesque Do you maybe know a person who has a merge rights and could review this PR? Could I help somehow with merging this PR?

bethesque:
Sorry, I maintain the ruby version, and don't know any python. @matthewbalvanz-wf is the one to assist.

bethesque:
One of the tests seems to be failing - can you see if you can get it green?

yangineer:
The pr that makes the tests green is in #95

DannyCork:
@matthewbalvanz-wf can this be approved ?

yangineer:
All the tests are passing after merging PR #95.
This PR is ready to be merged if anyone's free to review it.

matthewbalvanz-wf:
Sorry for the delay @yangineer thanks for the changes!

PR #95, opened 2018-10-20
 Updated the tests to be compatible with Click v7

yangineer:
Click 7x has changed the API for Result.
This fixes the tests and pins version 7 of Click.

FYI: https://click.palletsprojects.com/en/7.x/api/#click.testing.Result

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -109,7 +109,7 @@ def read(filename):
 
 
 dependencies = [
-    'click>=2.0.0',
+    'click>=7.0.0',
     'psutil>=2.0.0',
     'requests>=2.5.0',
     'six>=1.9.0',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -109,7 +109,7 @@ def read(filename):
 
 
 dependencies = [
-    'click>=2.0.0',
+    'click>=7.0.0',
```
cxong: Should this include a max version i.e. `click>=7.0.0,<8` to prevent surprise API breaks in the future?

yangineer: Can’t tell in advance if the tests will be affected by new breaking changes.
Don’t know if we want to limit ourselves to a specific version at this time.
If v8 comes out and updating is not worth the effort, then setting a max makes sense.

yangineer:
Merging this PR with #94.

PR #96, opened 2018-11-08
Add pypi badge to readme

cxong:


xen:
What is profit of using badges? Looks like modern search spam link building inside Github.

cxong:
@xen https://medium.freecodecamp.org/how-to-use-badges-to-stop-feeling-like-a-noob-d4e6600d37d2
Badges are a popular method of providing concise, quick and up-to-date info on a project. You'll note that this project already uses a few badges. This one provides the following benefits:

- shows the project is on pypi
- shows the pypi name at the top (granted, you could also infer the same info by reading down a bit to the `pip install pact-python` part
- shows the latest version published to pypi
- gives a link for people to go directly to the pypi project page

And all this in a tiny image that can be read at a glance. Try that with plain text.

I don't know what "search spam link" is, but badges are not. Pypi is not spam.

DannyCork:
can this be approved ? @cxong

PR #120, opened 2020-01-19
Pact provider options

elliottmurray:
Added pact broker, provider and consumer tag options and pass through. I suspect I haven't fully understood the consumer tag options so could do with some help understanding how to test this properly.


pact/verify.py
```diff
@@ -46,6 +46,15 @@
 @click.option(
     'username', '--pact-broker-username',
     help='Username for Pact Broker basic authentication.')
+@click.option(
+    'broker_base_url', '--pact-broker-base-url',
+    default='',
+    help='Base URl for the Pact Broker instance to publish pacts to.')
+@click.option(
```
bethesque: This needs to be able to be specified multiple times (eg. I want to verify "master" and "prod" pacts)

pact/verify.py
```diff
@@ -46,6 +46,15 @@
 @click.option(
     'username', '--pact-broker-username',
     help='Username for Pact Broker basic authentication.')
+@click.option(
+    'broker_base_url', '--pact-broker-base-url',
```
bethesque: If I was writing the pact CLI again, I'd just call this "pact-broker-url". I don't think it needs the "base". How about we drop it in the python?

pact/verify.py
bethesque: Or, allow a single `--consumer-version-tags` with multiple quotation delimited arguments, `--consumer-version-tags master prod "my feat"`

Same for the provider version tags.

pact/verify.py
bethesque: If I was writing the pact CLI again, I'd just call this "pact-broker-url". I don't think it needs the "base". How about we drop it in the python?

pact/test/test_verify.py
```diff
@@ -138,6 +178,40 @@ def test_all_options(self):
...
+            '--provider=provider',
+            '--publish-verification-results',
+            '--provider-app-version', '1.2.3',
+            '--verbose')
+
+    def test_all_broker_options(self):
+        self.mock_Popen.return_value.returncode = 0
+        result = self.runner.invoke(verify.main, [
+            '--pact-broker-base-url=http://localhost/broker',
+            '--consumer-version-tag=$LATEST',
```
bethesque: What is $LATEST here? Unlike docker, using the tag "latest" with the Pact Broker is confusing, as then you'd be retrieving "the latest pact with the tag 'latest'" eg. `/pacts/provider/my-provider/consumer/my-consumer/latest/latest`

bethesque: This is a good resource to read to get an idea of how tags work btw https://github.com/pact-foundation/pact_broker/wiki/Using-tags

pact/verify.py
```diff
@@ -125,6 +144,9 @@ def main(pacts, base_url, pact_url, pact_urls, states_url,
         '--provider-base-url': base_url,
         '--provider-states-setup-url': states_setup_url,
         '--broker-username': username,
+        '--pact-broker-base-url': broker_base_url,
+        '--provider': provider,
+        '--consumer-version-tag': consumer_version_tag,
```
bethesque: When the python interface has been updated to allow multiple tags,  `--consumer-version-tag` can be called once for each tag. Ditto for `--provider-version-tag`.

pact/verify.py
elliottmurray: Done

pact/test/test_verify.py
elliottmurray: Gotcha - so changed it to prod as it's just a unit test but I assume some clipped git sha would also do? $LATEST in this case was a lazy AWS reference on my part.

bethesque:
It's looking ok to me except for the multiple consumer and provider version tag support.

pact/test/test_verify.py
bethesque: You'd usually specify tag names, because you, as the provider, wouldn't necessarily know the exact version of the consumer in prod. The consumer team tells the pact broker when they deploy which version is in prod by tagging the application version resource, so the provider team can just say "the latest prod version" without having to know specifics. There's a good explanation here https://docs.pact.io/pact_broker/can_i_deploy

pact/verify.py
elliottmurray: I'm still a bit confused. I've tried reducing to the ruby gem to understand how this works currently and I don't get results I expect. So:

 pact-provider-verifier --provider-base-url=http://localhost:3000 --pact-broker-base-url=http://localhost:80 --provider=provider --consumer-version-tag=dev2 --consumer-version-tag=dev

Only calls one tag in broker (been monitoring requests) and doesn't actually verify. Well as far as I can tell. If I drop the tag it verifies (or at least I see that in the output)

elliottmurray:
@bethesque I think I have made the changes for multiple tags. Tests indicate desired behaviour (multiple tags tested for) I still have the problem sited around not getting the ruby gem working as I expect but that may be user error? I have left the docker file in there. I'll chat to you on Slack about what to do about that.

pact/verify.py
bethesque: I can't reproduce the multiple tag issue. You can have a play with the pact-provider-verifier docker container if that helps. 
The set up instructions are here: https://github.com/pact-foundation/pact-provider-verifier/blob/master/DEVELOPER_DOCUMENTATION.md

If you run `script/dev/broker.sh`, it will run against the test pactflow broker.

I'll try running your docker file.

bethesque: What do I do with the dockerfile to recreate the issue?

elliottmurray:
So I think my code copies the consumer tags behaviour correctly. However, I can't get 2 consumer tags to work against ruby (which means it doesn't work in python obviously) I have attached a screenshot - it just applies the last one. Hopefully something obvious? But I don't think I've introduced a regression into Python. 

Screen Shot 2020-02-04 at 10 04 23 pm

elliottmurray:
I believe I have the provider tags working too. An extra flag is a bit subtle. I might change that in a future PR and just rely on env vars.

elliottmurray:
Ah fixed it - using space instead of '=' for multiple key values 🤦‍♂

bethesque:
Ahh - that makes sense @elliottmurray. I'm not used to the = syntax (I always use spaces), and there was some custom handling around the repeated tags that I added. I'll make an issue to handle the = signs.

bethesque:
You can expose it as `--xxx-tags foo bar` in the python if that helps. I think I went the multiple parameter way because there's no rule that says that tags can't have spaces in them, but 99% of the tags I've seen don't have spaces, so it should be safe enough.

elliottmurray:
@bethesque I was thinking of that but life has got in the way. If I get a chance I'll create a separate PR to do it.

PR #106, opened 2019-06-11
Integrate pact broker client

mikahjc:
This adds a wrapper around the Ruby pact broker client that is pulled in with the Ruby mock service and provider verifier. It works similarly to how the mock service is started up. This would resolve issue #99.

PR #121, opened 2020-01-28
Updates to integrated pact broker client

mikahjc:
Minor fixes that didn't get pushed before #106 got merged.
* `publish()` method was looking to see if `self.broker_username` was set instead of `self.broker_password` when adding the `--broker-password` argument
* Documentation updated to reflect environment variables picked up by pact broker client
* Allow `publish()` to proceed if no pact broker base URL is provided, but `PACT_BROKER_BASE_URL` environment variable is set
* Add test regarding above code change


bethesque:
Looks good! I'll take the liberty of merging this.

Issue #111, opened 2019-09-27
Dynamic args to setup() is incompatible with PyCharm


antonagestam:
PyCharm complains that `Package containing module 'pact' is not listed in project requirements` when importing the pacts module. I'm guessing this is due to `setup()` being called with an expanded dict that PyCharm isn't looking at. Is there any specific reason for this, or can it be changed to a call with kwargs instead?

https://github.com/pact-foundation/pact-python/blob/master/setup.py#L121

elliottmurray:
Like so?
https://github.com/pact-foundation/pact-python/pull/124

PR #124, opened 2020-02-01
Converting to kwargs

elliottmurray:
Attempt to convert to kwargs in setup for Pycharm

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,28 +118,27 @@ dependencies = [
 if sys.version_info.major == 2:
     dependencies.append('subprocess32')
 
-setup_args = dict(
-    cmdclass={'develop': PactPythonDevelopCommand,
-              'install': PactPythonInstallCommand},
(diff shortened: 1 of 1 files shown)


antonagestam:
@mefellows This hopefully solves #111. There isn't anything programatically wrong with building the dict and then calling the function with its expansion. But, my guess is that this prevents PyCharm from statically parsing the file. So this is more about adhering to the current limits/design problems of Python packaging.

I'll try to get the time to test this out when I have the time!

mefellows:
Ah cool, that makes sense. Carry on :)

bethesque:
I totally know what a KWARGS are and I KWARG all the time, this is great KWARGING, there should be more of it, hooray for KWARGS!

bethesque:
Given I have NFI what this is about, I'll just see if other other Matt wants to weigh in before we merge this.

PR #129, opened 2020-04-01
Docker support for builds

elliottmurray:
Created some simple docker files and script to help create envs where not all python versions are available. Needs a bit of work but allows debugging of version issues. I could tweak this to publish containers with pact-python installed and published to pact-foundation?

### DIFF
--- a/docker/py36.Dockerfile
+++ b/docker/py36.Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.6.10-alpine3.11
+
+WORKDIR /home
+
+COPY requirements_dev.txt .
+
+RUN apk add gcc py-pip python-dev libffi-dev openssl-dev gcc libc-dev make
+
+RUN python -m pip install psutil subprocess32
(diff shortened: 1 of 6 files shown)


mefellows:
I'll try and have a look at running this later today Elliot.

docker/py36.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.6.10-alpine3.11
+
+WORKDIR /home
+
+COPY requirements_dev.txt .
+
+RUN apk add gcc py-pip python-dev libffi-dev openssl-dev gcc libc-dev make
+
+RUN python -m pip install psutil subprocess32
```
matthewbalvanz-wf, line 9: It's been a while, but I was thinking we don't need to install `subprocess32` outside of Python 2.7 as it's a backport of the Python 3.2+ version of subprocess.

elliottmurray: Removed. I'' add a commit with bash installed too

PR #130, opened 2020-04-03
Examples for python

elliottmurray:
Created a simple e2e example based on a few internet examples. Idea is to help people get started quicker with a reference implementation.


examples/e2e/provider.py
```diff
@@ -0,0 +1,49 @@
+import datetime
+
+from flask import Flask, abort, jsonify, request
+
+fakedb = {}
+
+app = Flask(__name__)
+
+
+@app.route('/_pact/provider_states', methods=['POST'])
```
mefellows, line 10: Is there a way you can dynamically add this endpoint during the verification test so it's clear this endpoint doesn't need to be present in a deployed environment? This might confuse some people coming in - we want to make it clear that the states endpoint should only be added for testing purposes.

mefellows:
Looks like a good start Elliot!

examples/e2e/README.md
```diff
@@ -0,0 +1,91 @@
+# Introduction
```
mefellows, line 1: Should we create a section in the main readme that points to these examples?

elliottmurray: Done

examples/e2e/provider.py
elliottmurray: I've been moving stuff from the e2e folder into examples (and from other sources) to make it a bit simpler for newbies. The current e2e tests are for local and weren't totally clear how to run. But I'll keep chipping away. I think I can do what you want via some kinda mix in in python but I'll look into it.

elliottmurray: I've created a first pass at this. There is in the root folder a pact_provider file that imports from the provider plus decorating with extra pact endpoints

elliottmurray:
@mefellows I think I have it integrated into CI/CD. Only runs on 3.6 - current one only does 2.7.

On that I think we should also update the versions? I'm tempted to drop support for 2.7 and we should definitely add 3.8. I can do the latter in a separate PR but leave the v2 as a separate issue.

PR #132, opened 2020-04-17
Move tests to standard tests dir

pyasi:
Moving the `test` dir that was previously part of the `pact` package into it's own module `tests`. This follows standard practice in python projects.

### DIFF
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -2,7 +2,7 @@ from unittest import TestCase
 
 from mock import patch
 
-from .. import constants
+from pact import constants as constants
 
 
 class broker_client_exeTestCase(TestCase):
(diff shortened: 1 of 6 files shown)


mefellows:
Seems like a good idea. Is there a reason why it may not have been there to begin with?

pyasi:
Very possible @mefellows, hoping for feedback if that’s the case. Diving into the code and it was something that caught my eye. Given it’s a small, low risk change i decided to throw this up for thoughts.

pyasi:
@mefellows Think this could get reviewed/merged? I'd be interested in continuing to contribute to Pact

mefellows:
Hi @pyasi, thanks and sorry for the delay. The core maintainer has been pretty flat out recently, so we're just waiting on another person (who knows what they're doing in the Python space) to finish the review.

If you're not already, join us at slack.pact.io and we can chat there - we'd love more contributions!

elliottmurray:
Sorry for the delay @pyasi but going in now

PR #135, opened 2020-04-30
Fix missing normalization of consumer name while publishing pact

m-aciek:
We have a name of consumer which is a string formatted as title: "Sample Project". It's contract file is being saved as "sample_project_….json". While publishing pact logic is looking for a name of contract file beginning with not normalized, but original consumer name. Therefore it does not find the file and exits with following message:

    No value provided for required pact_files
    Error in atexit._run_exitfuncs:
      File "/…/site-packages/pact/pact.py", line 260, in stop_service
        self.publish()
      File "/…/site-packages/pact/pact.py", line 193, in publish
        .format(url))
    RuntimeError: There was an error while publishing to the pact broker at ….

The pull request is a fix for that.

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -149,6 +149,10 @@ class Pact(object):
         self._interactions[0]['provider_state'] = provider_state
         return self
 
+    @staticmethod
+    def _normalize_consumer_name(name):
+        return name.lower().replace(' ', '_')
+
     def publish(self):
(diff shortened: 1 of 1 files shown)


elliottmurray:
Looks good - the only thing is changing unit tests? In test_pact.py maybe tweak the mock on line 229 to not be a mock? And then change the expectation below that in e.g. test_default_publish so we are exercising that bit of code.

PR #138, opened 2020-05-14
Pyasi add Matcher support for common regex formats

pyasi:
# Matching Common Formats
Add in support for matchers to match against common regex formats. This PR adds a class to the `matchers.py` file that can be imported and used to replace re-writing common regex patterns for standard formats. Adds unit tests to keep 100% coverage on the repository. I also have made changes to the E2E tests to incorporate a couple of examples.

 I've added support for:
- Integers
- Decimals
- Ids (identifiers)
- Hexadecimals
- IP addresses
- IPv6
- Dates
- Timestamps
- Times
- Uuids

### Usage example
```
from pact.matchers import Format
Format().integer  # Matches if the value is an integer
Format().ip_address  # Matches if the value is a ip address
```


elliottmurray:
I think you need to make a change to the pact/__init__.py file. Should be obvious but my local install doesn't find Format otherwise.

elliottmurray:
Version number needs updating to match. It probably will be ok on a merge but would prefer if you could rebase to latest master.

elliottmurray:
We need to think about the enum dependency. I don't think this will work on 2.7 as is because the package won't have it imported (just the CI/CD pipeline). It might need to have a line added to setup.py that itself might need a little work.

pyasi:
Comments addressed, good call!

Issue #141, opened 2020-06-02
can not mock a https server 


NickWang23:
I have post the detail on the stack overflow:
https://stackoverflow.com/questions/62144092/how-to-mock-a-https-server-by-pact-python-could-you-give-me-a-example
If you have anytime, please check this, thank you.

elliottmurray:
I can recreate locally and will investigate but when I think about it not totally surprised it is failing. Following on from the original reply to your S.O. post - can you just run it as http?

NickWang23:
I think I get your point, you mean we just need to focus on the content of contract and https is not a main aspect. Cuz I'm a rookie of contact test, I strict follow the rule: 100% simulation then replay. Thank you for your reply.

elliottmurray:
I've actually pushed 1.0.1 which does turn verification off. I don't think it does any real harm. But yes I think strictly it's unncecessary. I'll close this off and you're free to use 1.0.1 which should get you going.

PR #142, opened 2020-06-02
Allowed https communication to mock. Didn't fix tests

elliottmurray:
#141 feature request


mefellows:
Thanks Elliot. So it seems my suspicions were correct - it was Pact Python not ignoring certs!

Thanks for the fix.

PR #145, opened 2020-06-11
Remove typo from examples/e2e requirements.txt

jstoebel:
A stray f was in examples/e2e/requirements.txt causing pip install to break


elliottmurray:
Thanks! I need to work on these a bit more and incorporate them into the CI pipeline.

PR #147, opened 2020-06-24
Added logging params

elliottmurray:
Added logging params


elliottmurray:
@bethesque quick review please and I'll release. I have log dir working locally.

PR #150, opened 2020-06-26
ci: add check for commit messages

elliottmurray:
Testing git message script

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ rvm: "2.2"
 
 sudo: false
 
+before_install:
+  - script/commit_message.py
+
 install:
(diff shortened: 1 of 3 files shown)


elliottmurray:
@bethesque @mefellows I've made a small change to docs around contributing. Also added a pre-script step that checks the commits in a PR. Can you have a quick gander? I'm not totally sure I should do this!

mefellows:
I think it's worth a shot.

I've also just enabled https://github.com/zeke/semantic-pull-requests on the repo - which will update the status also. We can see which method is preferable (it might be preferable to have the build pass, but fail another status check - I'm not sure).

We use it for JS/Go and it's been helpful.

PR #160, opened 2020-07-22
docs: Add to README to show simple publish to broker"

elliottmurray:
Simple example to show publish in python code. WIP to show it via pact tooling

@bethesque just a quick fix to try and give some peeps an example

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,29 @@ Like({
 })
 ```
 
+## Publish to Broker
+The support in pact-python for publishing to broker is not fully supported currently. A sample method of how you might publish in python using requests follows:
+
+```python
+def push_to_broker(version):    
+    with open(os.path.join(PACT_DIR, PACT_FILE), 'rb') as pact_file:
+        pact_file_json = json.load(pact_file)
+
+    basic_auth = HTTPBasicAuth(PACT_BROKER_USERNAME, PACT_BROKER_PASSWORD)
+
+    log.info("Uploading pact file to pact broker...")
+
+    r = requests.put(
+        "{}/{}".format(PACT_UPLOAD_URL, version),
+        auth=basic_auth,
+        json=pact_file_json,
+        verify=False
+    )
+    if not r.ok:
+        log.error("Error uploading: %s", r.content)
+        r.raise_for_status()
+```
+
 For more information see [Matching](https://docs.pact.io/getting_started/matching)
 
 ## Verifying Pacts Against a Service


elliottmurray:
@mefellows this one? https://github.com/pact-foundation/pact_broker-client

I think it is but didn't have some docs to hand. I'll look it up - probably better to use that. I think Python wraps it but would have to look closer at the code.

elliottmurray:
I'm looking at the setup code and I'm not sure it is now. The pact-ruby-standalone is (and what is used for verify) Is it bundled in there? I can see a publish results in that method but I think I know what that is and it isn't push to broker. There are placeholders for it and I think there was some sample code but not sure it was ever completed. 

What do you want me to do? I can just link to the cli instead of what I have? Might be better than getting people to do custom PUTS?

bethesque:
My original thought was to just create a top level command that delegates everything (no validation, just pass on the args) to the standalone. That would be easiest and the most maintainable.

bethesque:
No, we definitely don't want people doing manual puts. The ordering of the tag and pact creation is critical for the webhooks to fire correctly, so we don't want people doing it manually.

mefellows:
Sorry Elliot I should have been clearer last night.

My understanding is that when pact-python is installed, it pulls down the [standalone](https://github.com/pact-foundation/pact-ruby-standalone) CLI tool suite, which is used to power the project.

They get bundled in to the pip install path e.g.

```
➜  e2e git:(master) ✗ ll /usr/local/lib/python3.7/site-packages/pact/bin/pact/bin
total 28K
-rwxr-xr-x 1 matthewfellows admin 958 Aug 15  2019 pact*
-rwxr-xr-x 1 matthewfellows admin 965 Aug 15  2019 pact-broker*
-rwxr-xr-x 1 matthewfellows admin 966 Aug 15  2019 pact-message*
-rwxr-xr-x 1 matthewfellows admin 971 Aug 15  2019 pact-mock-service*
-rwxr-xr-x 1 matthewfellows admin 993 Aug 15  2019 pact-provider-verifier*
-rwxr-xr-x 1 matthewfellows admin 984 Aug 15  2019 pact-publish*
-rwxr-xr-x 1 matthewfellows admin 971 Aug 15  2019 pact-stub-service*
drwxr-xr-x 9 matthewfellows admin 288 Aug 15  2019 ./
drwxr-xr-x 5 matthewfellows admin 160 Aug 15  2019 ../
```

In Pact Node, for example, we create bin stubs (essentially aliases that show up in `./node_modules/bin` that just point at these scripts). Super convenient, because you can just run `pact-broker publish ...` from the context of the NodeJS project and viola. And you don't need to wrap them in NodeJS code.

Is there something equivalent we can do in Pact Python?

elliottmurray:
not needed

PR #171, opened 2020-09-03
feat: Support pending Pacts

### LINKED ISSUE #164: Support pending pacts
Hi,
Please add support for the new "pending pacts" feature:
https://pact.canny.io/feature-requests/p/pending-pacts-prevent-consumers-being-able-to-break-provider-builds-with-changed

I might be able to open a PR for this myself if I get some guidance.

m-aciek:
Closes #164.

### DIFF
--- a/pact/cli/verify.py
+++ b/pact/cli/verify.py
@@ -110,10 +110,17 @@ import click
 @click.option(
     'log_level', '--log-level',
     help='The logging level.')
+@click.option(
+    'enable_pending', '--enable-pending',
+    default=False,
+    help='Allow pacts which are in pending state to be verified without causing the '
+         'overall task to fail. For more information, see https://pact.io/pending',
+    is_flag=True)
 def main(pacts, base_url, pact_url, pact_urls, states_url, states_setup_url,
          username, broker_base_url, consumer_version_tag, consumer_version_selector,
          provider_version_tag, password, token, provider, headers, timeout,
-         provider_app_version, publish_verification_results, verbose, log_dir, log_level):
+         provider_app_version, publish_verification_results, verbose, log_dir,
+         log_level, enable_pending):
     """
     Verify one or more contracts against a provider service.
 
@@ -168,7 +175,7 @@ def main(pacts, base_url, pact_url, pact_urls, states_url, states_setup_url,
         'consumer_tags': list(consumer_version_tag),
         'consumer_selectors': list(consumer_version_selector),
         'provider_tags': list(provider_version_tag),
-        'provider_states_setup_url': states_setup_url
+        'provider_states_setup_url': states_setup_url,
     }
 
     options = dict(filter(lambda item: item[1] is not None, options.items()))
@@ -178,6 +185,7 @@ def main(pacts, base_url, pact_url, pact_urls, states_url, states_setup_url,
     success, logs = VerifyWrapper().call_verify(*all_pact_urls,
                                                 provider=provider,
                                                 provider_base_url=base_url,
+                                                enable_pending=enable_pending,
                                                 **options)
     sys.exit(success)
 
--- a/pact/cli/verify.py
+++ b/pact/cli/verify.py
@@ -105,10 +105,6 @@ import click
     default=False,
     help='Toggle verbose logging, defaults to False.')
 @click.option(
-    '--verbose/--no-verbose',
-    default=False,
-    help='Toggle verbose logging, defaults to False.')
-@click.option(
     'log_dir', '--log-dir',
     help='The directory for the pact.log file.')
 @click.option(
--- a/pact/verify_wrapper.py
+++ b/pact/verify_wrapper.py
@@ -133,7 +133,9 @@ class VerifyWrapper(object):
 
         return True
 
-    def call_verify(self, *pacts, provider_base_url, provider, **kwargs):
+    def call_verify(
+            self, *pacts, provider_base_url, provider, enable_pending=False, **kwargs
+    ):
         """Call verify method."""
         verbose = kwargs.get('verbose', False)
 
@@ -169,6 +171,8 @@ class VerifyWrapper(object):
 
         if(kwargs.get('verbose', False) is True):
             command.extend(['--verbose'])
+        if enable_pending:
+            command.append('--enable-pending')
 
         headers = kwargs.get('custom_provider_headers', [])
         for header in headers:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/commitizen-tools/commitizen
+    rev: master
+    hooks:
+      - id: commitizen
+        stages: [commit-msg]
--- a/pact/verifier.py
+++ b/pact/verifier.py
@@ -94,7 +94,6 @@ class Verifier(object):
         provider_app_version = kwargs.get('provider_app_version', None)
         headers = kwargs.get('headers', [])
         timeout = kwargs.get('timeout', None)
-        verbose = kwargs.get('verbose', None)
         consumer_tags = kwargs.get('consumer_tags', [])
         provider_tags = kwargs.get('provider_tags', [])
         states_setup_url = kwargs.get('provider_states_setup_url', None)
@@ -107,7 +106,6 @@ class Verifier(object):
             'provider_app_version': provider_app_version,
             'custom_provider_header': list(headers),
             'timeout': timeout,
-            'verbose': verbose,
             'consumer_tags': list(consumer_tags),
             'provider_tags': list(provider_tags),
             'provider_states_setup_url': states_setup_url,
--- a/pact/verifier.py
+++ b/pact/verifier.py
@@ -97,6 +97,7 @@ class Verifier(object):
         states_setup_url = kwargs.get('provider_states_setup_url', None)
         verbose = kwargs.get('verbose', False)
         publish_version = kwargs.get('publish_version', None)
+        enable_pending = kwargs.get('enable_pending', None)
 
         options = {
             'log_dir': log_dir,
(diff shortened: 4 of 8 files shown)


matthewbalvanz-wf:
@m-aciek `mock` should be part of the Python standard library from 3.3 and above, maybe we could remove that entirely now that support for Python 2.7 has been removed?
https://docs.python.org/3/library/unittest.mock.html

m-aciek:
@matthewbalvanz-wf Unfortunately in 3.4 and 3.5 mock that is shipped with unittest has different API than in 3.8 so the asserts in VerifyWrapper fail.

pact/verify_wrapper.py
```diff
@@ -169,6 +169,7 @@ def call_verify(self, *pacts, provider_base_url, provider, **kwargs):
 
         if(kwargs.get('verbose', False) is True):
             command.extend(['--verbose'])
+        command.extend(self._get_enable_pending_flag(kwargs.get('enable_pending')))
```
przemyslaw-jakielaszek, line 172: _get_boolean_flag('pending')

pact/verify_wrapper.py
```diff
@@ -169,6 +169,7 @@ def call_verify(self, *pacts, provider_base_url, provider, **kwargs):
 
         if(kwargs.get('verbose', False) is True):
```
przemyslaw-jakielaszek, line 170: command.extend(self. _get_boolean_flag('verbose'))

przemyslaw-jakielaszek (changes requested):

pact/cli/verify.py
```diff
@@ -104,20 +104,23 @@
...
-    default=False,
-    help='Toggle verbose logging, defaults to False.')
 @click.option(
     'log_dir', '--log-dir',
     help='The directory for the pact.log file.')
 @click.option(
     'log_level', '--log-level',
     help='The logging level.')
+@click.option(
+    'enable_pending', '--enable-pending/--no-enable-pending',
```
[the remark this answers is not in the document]
m-aciek: I wanted to replicate the flag that is used in underlying CLI and to be in line with flag name used in documentation https://docs.pact.io/pact_broker/advanced_topics/pending_pacts/#to-start-using-the-pending-pacts-feature.

I used default=None also to replicate usage of underlying CLI so to map False -> --enable…, True -> --no-enable…, None -> no flag.

pact/verify_wrapper.py
```diff
@@ -199,3 +200,11 @@ def publish_results(self, provider_app_version, command):
         command.extend(["--provider-app-version",
                        provider_app_version,
                        "--publish-verification-results"])
+
+    @classmethod
+    def _get_enable_pending_flag(cls, flag_value):
```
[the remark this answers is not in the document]
m-aciek: There is no such a flag as --no-verbose, so I think it is a premature optimization.

pact/verify_wrapper.py
m-aciek: >There is no such a flag as --no-verbose, and for just one case I think it is a premature optimisation.

pact/verify_wrapper.py
m-aciek: >There is no such a flag as --no-verbose, and for just one case I think it is a premature optimisation.

m-aciek:
@matthewbalvanz-wf Summary of failing tests because of mock in 3.5 compatibility issues:

```
Error
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 605, in run
    testMethod()
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 1159, in patched
    return func(*args, **keywargs)
  File "/Users/maciek/projects/pact-python/tests/cli/test_verify.py", line 122, in test_broker_url_and_provider_required
    mock_wrapper.assert_called()
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 585, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called

Error
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 605, in run
    testMethod()
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 1159, in patched
    return func(*args, **keywargs)
  File "/Users/maciek/projects/pact-python/tests/cli/test_verify.py", line 339, in test_expand_directories_called
    mock_expand_dirs.assert_called_once()
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 585, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once

Error
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 605, in run
    testMethod()
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 1159, in patched
    return func(*args, **keywargs)
  File "/Users/maciek/projects/pact-python/tests/test_verifier.py", line 74, in test_expand_directories_called_for_pacts
    mock_expand_dir.assert_called_once()
  File "/Users/maciek/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 585, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once

SubTest failure: Traceback (most recent call last):
  File "/…/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/…/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 523, in subTest
    yield
  File "/…/projects/pact-python/tests/test_verifier.py", line 84, in test_passes_enable_pending_flag_value
    mock_wrapper.call_args.kwargs,
  File "/…/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 682, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true : kwargs

Error
  File "/…/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/…/.pyenv/versions/3.5.9/lib/python3.5/unittest/case.py", line 605, in run
    testMethod()
  File "/…/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 1159, in patched
    return func(*args, **keywargs)
  File "/…/pact-python/tests/test_verify_wrapper.py", line 160, in test_rerun_command_called
    mock_rerun_cmd.assert_called_once()
  File "/…/.pyenv/versions/3.5.9/lib/python3.5/unittest/mock.py", line 585, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once
```

matthewbalvanz-wf:
Looks like `assert_called` and `assert_called_once` were added in 3.6, so it seems like using the library until that is end of life is the path forward.

pact/verify_wrapper.py
przemyslaw-jakielaszek: Maybe I am wrong, but I think there is a `no-verbose` flag.
https://github.com/pact-foundation/pact-python/pull/171/files#diff-99dbf564da4f139a45a61bf9c28b4f28R104

pact/cli/verify.py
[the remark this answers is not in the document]
przemyslaw-jakielaszek: 1. you are right. It is said to use `enable-pending` convention: https://github.com/pact-foundation/pact_broker/issues/320
2. it is also said that flag should be set to False as a default value. Moreover I do not see a case when `no flag` differs from `no-enable-pending`

przemyslaw-jakielaszek (changes requested):

pact/cli/verify.py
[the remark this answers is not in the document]
m-aciek: Ok, so to be in line with the guidelines you linked and to make the implementation minimal, I will stick only with positive flag, set False as default and do not add --no-enable-pending to underlying library.

pact/verify_wrapper.py
m-aciek: It is lacking in the ruby verifier help message (see https://github.com/pact-foundation/pact-provider-verifier/blob/master/README.md), but I checked and indeed this ruby binary accepts --no-verbose. Although according to https://github.com/pact-foundation/pact-python/pull/171#discussion_r483937714 I'll drop the negative flag passing.

m-aciek:
I've added Python 3.4 syntax compatibility and simplified code. Ready for review again.

m-aciek:
I've fixed code issue with Click option.

bethesque:
Would you mind leaving it as --enable-pending and --no-enable-pending in the python interface? When trying to write documentation that is applicable for all languages, it makes it very difficult when each of the 11 languages chooses a slightly different option name!

elliottmurray:
@bethesque isn't the 'no' option redundant? Just wondering if it should be default off and you just ask --enable-pending but not sure what the plans are for this ongoing?

m-aciek:
@bethesque @elliottmurray I had the same doubt and simplified implementation to avoid redundancy.

The use case for --no-enable-pending that I can think of would be having some default configuration that could be overridden in a command call.

I plan to implement support for --no-enable-pending on another branch and open new alternative pull request to avoid going round in circles in case of further discussion.

m-aciek:
@bethesque @elliottmurray 

> The use case for --no-enable-pending that I can think of would be having some default configuration (e.g. in pyproject.toml) that could be overridden in a command call.

In this use case default for enable-pending flag would be `None`, meaning „not overriding some default configuration” in case of no of those flags are added. I'm assuming this in the implementation.

elliottmurray:
I'm going to merge this and do a release later. It doesn't harm as far as I can tell and we can add the extra flag in another PR

bethesque:
Sorry, I don't care about --no-enable-pending, just the --enable-pending (rather than --pending or whatever it was)

PR #172, opened 2020-09-09
feat(verifier): support --no-enable-pending flag in CLI

m-aciek:
This pull request implements following behaviour:
* for --enable-pending flag in CLI and enable_pending=True in Python API it adds --enable-pending in underlying CLI,
* for --no-enable-pending flag in CLI and enable_pending=False in Python API it adds --no-enable-pending in underlying CLI,
* for lack of those flags in CLI and enable_pending=None (default for optional keyword argument) it doesn't add any of those flags in underlying CLI.

### DIFF
--- a/pact/cli/verify.py
+++ b/pact/cli/verify.py
@@ -111,11 +111,12 @@ import click
     'log_level', '--log-level',
     help='The logging level.')
 @click.option(
-    'enable_pending', '--enable-pending',
+    'enable_pending',
+    '--enable-pending/--no-enable-pending',
     default=False,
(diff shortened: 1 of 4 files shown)


bethesque:
Sorry, as commented on the other issue, I don't care about the --no-enable-pending. I don't know if I'm going crazy, but I swear when I looked at the code the enabled option wasn't `--enable-pending`, it was something else. As long as the positive case is --enable-pending, that's all I care about. Thanks!

m-aciek:
Review comments might have mislead you. No worries. :) Thank you for the comment.

Therefore I'm closing this pull request.

PR #173, opened 2020-09-10
Allow to use unauthenticated brokers

copalco:
There is a bug in a pact verifier which  prohibits use of unauthenticated pact-broker


elliottmurray:
Can you add some asserts around line 87-ish. There are some examples in that file but at the moment you don't seem to actually assert anything. Otherwise LGTM. 

Though I didn't actually know you could use broker without auth?

copalco:
@elliottmurray I don't assert anything, because this test was raising an error, before I applied the change, so we are implicitly asserting that it's working fine now.

copalco:
@elliottmurray Yes we have a broker without authentication configured and it's working fine.

elliottmurray:
I accept that fixes this immediate problem but the test should describe the expected behaviour that we make a process call with this in place. See line 138. You should be able to copy/paste with a small modification or 2.

copalco:
@elliottmurray I added assertion that popen process was called with expected parameters

elliottmurray:
Thanks

PR #182, opened 2020-10-17
feat(verifier): Support work-in-progress pacts attribute in CLI

thatguysimon:
Adds a wrapper around the `include-wip-pacts-since` attribute in the CLI


thatguysimon:
Hi @elliottmurray,
Currently the build is failing due to flake8's:
```
./pact/verify_wrapper.py:132:5: C901 'VerifyWrapper.call_verify' is too complex (11)
```
I guess the new condition I added was one too many for flake8.

Is there a particular way you would like me to solve this?

One quick way that I can think of to make the function less complex is to extract the following piece into a separate "validation" function:
```python
if(len(pacts) == 0 and not self._broker_present(**kwargs)):
    raise PactException('Pact urls or Pact broker required')
```

elliottmurray:
I think something like that is acceptable. I probably need to revisit that method at some point or exclude it from flake as I've hit this a bunch of times. But for your PR that is totally fine

thatguysimon:
Done.
Let me know if there's anything else I should change in the PR @elliottmurray

elliottmurray:
That's great - I'll do a release later or tomorrow. Thanks

PR #189, opened 2020-12-04
docs: typo in pact-verifier help string: PUT -> POST

noelslice:
`pact-verifier` sends POST requests when setting up the provider states.


elliottmurray:
Thanks!

PR #191, opened 2020-12-17
ci(github actions): added Github Actions configuration for build and test

### LINKED ISSUE #190: chore: migrate away from TravisCI
TravisCI (org) is dropping support for OSS builds at the end of this month, meaning they must be transferred to travis-ci.com, which uses a credit system. Builds may fail (and already are) if the number of credits is exhausted within th emonth.

We are recommending switching to GH Actions, but the choice of CI provider is up to the project maintainer.

See also:

* https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom
* https://travis-ci.community/t/org-com-migration-unexpectedly-comes-with-a-plan-change-for-oss-what-exactly-is-the-new-deal/10567/5

matthewbalvanz-wf:
Created a configuration that runs lints and tests on each push to Github. Removed Python 2.7 as an
environment as current versions of Pip are now correctly noting we are trying to install a version
of pydocstyle that does not support 2.7.

resolves #190
 
I have not moved the parts of the Travis CI configuration that publishes to PyPi as it appears I lack the admin permissions to this repository to add credentials as a [secret for the build](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets). I believe we have been using my PyPi credentials in Travis CI though, so I think we either need to:

1. Get my or Elliot's credentials in Github as a secret for the build
2. Setup a "Pact" user in PyPi and add it as a project maintainer there and use it in Github instead

@bethesque @mefellows I suspect one of you may be admins in Github, do you have a preference on how we want to manage those publishing credentials?


mefellows:
Thanks for doing this! I've just made the pact python team Admins of this repo. You should be able to grab them now.

PR #195, opened 2021-01-19
fix: send to cli pact_files with the pact_dir in their path

cdambo:


elliottmurray:
@cdambo thanks for the contribution!

PR #200, opened 2021-01-26
Feature/message provider

tuan-pham:


mefellows:
Awesome Tuan!!

pact/http_proxy.py
```diff
@@ -0,0 +1,91 @@
+"""Http Proxy to be used as provider url in verifier."""
```
elliottmurray, line 1: Is the intention that Flask will run in the same process as out tests but in the background? Just beware you might have problems with later versions (Python and Flask) of this being a subprocess to do with tickle? Not a joke!

mefellows: > ..ave problems with later versions (Python and Flask) of this being a subprocess to do with tickle? Not a joke!
😆 

> Is the intention that Flask will run in the same process as out tests but in the background

In case the context isn't clear. Basically, the way the provider verification for messages works still uses the same underlying verification binary, which expects to talk to an HTTP server. Instead of making the end user do the mapping, this proxy server is responsible for mapping the "message" to be verified, to a "function" that will produce the message - i.e. the actual provider.

See the sequence diagrams here for more: https://github.com/pact-foundation/pact-message-demo

If you knew all that already... sorry!

tuan-pham: @elliottmurray yes, the idea is to run it as a subprocess. There maybe a problem with later version with subprocess here but I'm not 100% sure what can we replace a subproces with. 
@mefellows Thanks for reminding about sequence diagram. It helps understand the verify flow.

pact/message_pact.py
```diff
@@ -136,7 +136,7 @@ def with_content(self, contents):
         :rtype: Pact
         """
         self._insert_message_if_complete()
-        self._messages[0]['contents'] = contents
+        self._messages[0]['contents'] = from_term(contents)
```
tuan-pham, line 139: Well, we need test for this!

pact/http_proxy.py
elliottmurray: @mefellows I still don't understand this! I mean I get the provider side but the consumer side just makes my head hurt!

mefellows: That sequence diagram could do with some tuning.

Basically, on the consumer side of things the underlying Pact binaries aren't really involved in the testing - just serialisation of the contract (and reification of the message)

1. Consumer sets its expectations for what it can receive from a message provider and provides a way to invoke the message consumer (e.g. registering the handler function)
2. The framework is responsible for taking the input, stripping away the matchers etc. (reification) and then invoking the message consumer function with the expected input message
3. The framework checks if the message handler succeeds or fails. If it fails, the test ends. If it passes, it calls the binary to append the interaction to the contract

So basically, if the message handler doesn't fail, we consider the test successful. It's not quite the same as the HTTP pact, where the framework performs assertions on the HTTP request the consumer performs etc.

Does this clarify things at all?

tuan-pham: > I think you will need to use FastAPI over Flask @tuan-pham . I have an example in the examples. It is why I actually can't use the provider python in the e2e (which is Flask) and have to use a shell script instead.
@elliottmurray: Thanks, I will have a look into FastAPI!
Also, I understand that FLask is not production ready. However, I think it's perfect for the producer test as we can spin it up and tear down when done. What exact problem you have with Flask? 🤔

elliottmurray: Flask is certainly good enough for prod! But more that you may struggle with using a subprocess to manage it. If you don't need to do that use Flask.

elliottmurray: @mefellows thanks! Yes I do kind of get it with the biggest difference the http request. It's just a bit harder to get your head around.

antonakospanos:
Any progress here, will this PR be merged? We'd use PACT on RMQ consumers too.

pact/http_proxy.py
maggieleake: Hi I download @tuan-pham code.  I'm using python 3.7 and lastest greatest flask.  Which version of flask works?  The issue I'm seeing is in the setup we set values on localstack.  Then in the request we are getting None from the localstack.   Not sure how to fix this.  I figure different version of flask.

pulphix: > I think you will need to use FastAPI over Flask @tuan-pham . I have an example in the examples. It is why I actually can't use the provider python in the e2e (which is Flask) and have to use a shell script instead.

Hi @elliottmurray,
I'm working on the porting to FastAPI of the @tuan-pham work, any suggestions on how to replace werkzeug.local.LocalStack?

elliottmurray: > > I think you will need to use FastAPI over Flask @tuan-pham . I have an example in the examples. It is why I actually can't use the provider python in the e2e (which is Flask) and have to use a shell script instead.
> 
> Hi @elliottmurray,
> I'm working on the porting to FastAPI of the @tuan-pham work, any suggestions on how to replace werkzeug.local.LocalStack?

I've not looked at this recently due to personal stuff. I am going to start looking at moving to the underlying Rust impl and v3. However, I did do some work on this around messaging generally. I can't remember if I pulled Flask out on this branch or was just local:
https://github.com/pact-foundation/pact-python/commits/docs/kafka_example

The main issue IIRC was getting it to start up. But didn't get very far.

tuan-pham: @pulphix Yes, it's the pla to move to FastAPI over Flask. Unfortunately, it's not on my radar due to more higher priority assignment at my end. I may have a look into it at some point.

maggieleake: I setup maemcahce to solve the problem but fast api will solve this issue.

PR #203, opened 2021-02-23
fix: introduced and renamed specification version

elliottmurray:


elliottmurray:
Reviewed it and I think this is all good. We were using this param in our e2e anyway and it still works

Issue #221, opened 2021-04-12
Publishing verifier through CLI


olivzz:
Hi, I am using pact-python 1.3.4. Does pact-verify has the capability to stop from publishing or by default "publish-verification-results" seems to be false but still it is publishing to the broker. Moreover I am not able to set any values to the parameter as seen below I get error.
Error: --publish-verification-results option does not take a value

elliottmurray:
If you put a provider version number in it will by default try to publish. I did this thinking that people would want to do both all times? I can change it and add that flag if necessary. I think you might be able to override with the env var (something like PUBLISH_VERIFICATION_RESULTS=false)

olivzz:
I guess you mean provider_app_version. If so, can provider_app_version be left as optional? I tried but the test failed so I think it is a manadatory param.

elliottmurray:
#222 if you want to check it out and test it. It seems to be correct behaviour now. I'll create a release probably at the weekend

elliottmurray:
Released with 1.3.6

PR #222, opened 2021-04-14
fix: publish verification results was wrong

elliottmurray:
#221 issue


pact/verify_wrapper.py
```diff
@@ -165,9 +162,10 @@ def call_verify(
 
         if(provider_app_version):
             command.extend(["--provider-app-version",
-                            provider_app_version,
-                            "--publish-verification-results"])
-            # self.publish_results(provider_app_version, command)
+                            provider_app_version])
+
+        if kwargs.get('publish_verification_results', False):
```
mefellows, line 167: Do you need to explicitly check the value is True (as per the check for `verbose` on line 170)?. I must admit on first reading, I thought it was comparing to `False` which seemed backwards, but I think that's the default value.

mefellows (approved):

pact/verify_wrapper.py
elliottmurray: I think it works (per the unit tests) but I agree it's open to error and hard to read. I'll look at tweaking it.

Issue #228, opened 2021-04-27
Format().timestamp matcher usage results in TypeError


mefellows:
Repro code base: https://github.com/pactflow/example-consumer-python/blob/issue/matcher-not-serialisable/tests/consumer/test_products_consumer.py#L48

Steps to repro:

```sh
pip install -r requirements.txt
make test
```

Error:

```
======================================================================================= test session starts =======================================================================================
platform darwin -- Python 3.9.1, pytest-5.4.1, py-1.10.0, pluggy-0.13.1
rootdir: /Users/matthewfellows/development/public/example-consumer-python
collected 1 item

tests/consumer/test_products_consumer.py F                                                                                                                                                  [100%]

============================================================================================ FAILURES =============================================================================================
________________________________________________________________________________________ test_get_product _________________________________________________________________________________________

pact = <pact.pact.Pact object at 0x109ddc520>, consumer = <src.consumer.ProductConsumer object at 0x109ddc070>

    def test_get_product(pact, consumer):
        expected = {
            'id': "27",
            'name': 'Margharita',
            'type': 'Pizza',
            'date': Format().timestamp
        }

        (pact
         .given('a product with ID 10 exists')
         .upon_receiving('a request to get a product')
         .with_request('GET', '/product/10')
         .will_respond_with(200, body=Like(expected)))

>       with pact:

tests/consumer/test_products_consumer.py:57:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/local/lib/python3.9/site-packages/pact/pact.py:382: in __enter__
    self.setup()
/usr/local/lib/python3.9/site-packages/pact/pact.py:207: in setup
    resp = requests.put(
/usr/local/lib/python3.9/site-packages/requests/api.py:134: in put
    return request('put', url, data=data, **kwargs)
/usr/local/lib/python3.9/site-packages/requests/api.py:61: in request
    return session.request(method=method, url=url, **kwargs)
/usr/local/lib/python3.9/site-packages/requests/sessions.py:516: in request
    prep = self.prepare_request(req)
[...]

PR #230, opened 2021-04-27
fix: fix datetime serialization issues in Format

DawoudSheraz:
### Description

This PR fixes the usage of `Format().timestamp`, `Format().date`, and `Format().time` by converting the datetime objects into string isoformat. See https://github.com/pact-foundation/pact-python/issues/228

### DIFF
--- a/tests/test_matchers.py
+++ b/tests/test_matchers.py
@@ -362,7 +362,7 @@ class FormatTestCase(TestCase):
                         "s": self.formatter.Regexes.timestamp.value,
                         "o": 0,
                     },
-                    "generate": datetime.datetime(2000, 2, 1, 12, 30, 0, 0),
+                    "generate": datetime.datetime(2000, 2, 1, 12, 30, 0, 0).isoformat(),
                 },
             },
         )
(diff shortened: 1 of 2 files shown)


elliottmurray:
Merged - thanks for the contribution. I'll release probably Sunday but want to update the examples to use this first.

tests/test_matchers.py
```diff
@@ -400,7 +400,7 @@ def test_time(self):
                         "o": 0,
                     },
                     "generate": datetime.datetime(
-                        2000, 2, 1, 12, 30, 0, 0).time(),
+                        2000, 2, 1, 12, 30, 0, 0).time().isoformat(),
```
mefellows, line 403: Given that the string value should never change, should we put the expected string here? It's still not visible what the output would look like. I'd expect the output of isoformat for example to always produce the same formatted string (which I think is actually not correct for the other matcher types)

DawoudSheraz: As per https://github.com/DawoudSheraz/pact-python/blob/2403af78fb05c0a441da0da38b1bf43692239b6a/pact/matchers.py#L232, the generated values are static. I understand your point to add the explicit value in the tests(which I will do in a follow-up PR).

DawoudSheraz:
> Merged - thanks for the contribution. I'll release probably Sunday but want to update the examples to use this first.

Thank you.

PR #245, opened 2021-07-19
Feat/ffi provider

elliottmurray:
Draft PR for FFI Provider work


pact/ffi/ffi_verifier.py
```diff
@@ -0,0 +1,34 @@
...
+    def _load_ffi_library(self, ffi):
+        """Load the right library."""
+        target_platform = platform.platform().lower()
+
+        if 'darwin' in target_platform or 'macos' in target_platform:
+            libname = "libpact_ffi-osx-x86_64.dylib"
+        elif 'linux' in target_platform:
+            libname = "./libpact_ffi-linux-x86_64.so"
+        elif 'windows' in target_platform:
+            libname = "libpact_ffi-osx-x86_64.dylib"
```
uglyog, line 27: should be `.dll` for Windows

setup.py
```diff
@@ -56,6 +60,71 @@ def run(self):
...
+    else:
+        msg = ('Unfortunately, {} is not a supported platform. Only Linux,'
+               ' Windows, and OSX are currently supported.').format(
+            platform.platform())
+        raise Exception(msg)
+
+    if 'windows' in platform.platform().lower():
+        fetch_lib(bin_path, suffix, 'dll')
+        fetch_lib(bin_path, suffix, 'dll.lib')
+        fetch_lib(bin_path, suffix, 'lib')
```
uglyog, line 89: If you are using the DLL, you don't need the static lib (.lib). But there is no harm in downloading it, apart from it being the biggest one.

PR #251, opened 2021-07-28
feat: added support for message provider

pulphix:
I reworked the message provider support starting from the @tuan-pham branch and Rebase it.
I did update the commits comments to be compliant and fix the tests.

I changed as well the pact file naming removing the _message, checking other language implementation the _message was not in there and the Provider and Broker name with the _message will result alternated. The specification does not say that the queue contract file should contain _message


pulphix:
Hi @tuan-pham @elliottmurray,
I did update the test and now the integration tests are working as expected. the problem was that the contract generated by the consumer was not compliant with the producer messages, basically a typical example of integration issue that pact prevents :).
Please let me know if there are any other changes that I need to apply to the code.

elliottmurray:
If you have a chance before I release this can you come on Slack and DM me?

Issue #116, opened 2019-12-26
Problem installing pact standalone ruby app from github - Proxy Blocks URL


guilhermecgs:
Hi Folks, 

I am having problem downloading ruby standalone app from [github URL.](https://github.com/pact-foundation/pact-python/blob/master/setup.py#L64)

The company I work for blocks github url by default using a proxy. 

Can I make a PR to a setup.py that only python itself is installed (and do not try to install ruby app)?

(My plan is to install ruby app manually OR use a docker image...)

mefellows:
We have a similar facility in the JS implementation (see https://github.com/pact-foundation/pact-node/blob/master/README.md#pact-download-location).

Pact Go assumes it's on the path and checks it's installed correctly. You may need to include a check for the correct installed versions if you do this, because now the implementation can get out of sync with the dependencies.

guilhermecgs:
@mefellows : So, the work outline would be: (correct me if I am wrong)

1. Create a option to install only the python library and do not install the ruby app (in setup.py)
2. Create some sort of configuration to set the ruby app binaries location, just like "pact_binary_location" in JS implementation
3. When setting this configuration, check if the correct version is installed

mefellows:
I'm not familiar with idiomatic Python ways of doing these sorts of things, but yes something like the above sounds good.

Would love @matthewbalvanz-wf to chip in as he's the core maintainer here.

fresto32:
Hi All, is there a workaround for this? I'm running in an environment where I can only access github through a reverse proxy :(

mefellows:
I'm not sure of a workaround, but the ideal state I think is that Python bundles the Ruby application into the package, so at install time no network is required.

This is the plan for the FFI interface.

PR's welcome!

fresto32:
Thanks for the quick response Matt and a path forward!

YOU54F:
Thanks for the great contribution @taj-p closing this now as complete

PR #256, opened 2021-08-19
chore: Bundle all Ruby standalone binaries into wheel when deploying

fresto32:
This pull request modifies `setup.py` to allow for offline installation of this package.

In a corporate environment, it is very difficult to garner adoption of a product that is difficult to install in an offline environment. This PR hopes to smooth out that process.

Users can now install this package via `pip install pact-python --bin-path=/../../../` to prevent the installation process from trying to fetch binaries from github.com.

This seems to be a short term solution to https://github.com/pact-foundation/pact-python/issues/116. I am not sure whether the maintainers would want to close that issue or keep it open for the time being. As per @mefellows, "... the ideal state I think is that Python bundles the Ruby application into the package, so at install time no network is required." But this is planned for the FFI interface, which is still in development.

Also, this is my first time working within the pact-foundation, so please excuse any mistakes / issues I unintendedly may have made.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 
 import os
 import platform
+import shutil
 import sys
 import tarfile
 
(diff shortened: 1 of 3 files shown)


mefellows:
Thanks @fresto32 !

I think this PR is helpful, but I think a more helpful one would be to package the standalone into the pip package - is that possible?

With the proposed approach, you now need to sync an offline package (the standalone) with the pip package. Ideally, the whole package can be self contained.

fresto32:
Wow that was a quick turnaround. Thanks - yeah, if you're happy to bundle the current `PACT_STANDALONE_VERSION` standalones into the pip package, I'm more than happy to implement it! Stay tuned :)

mefellows:
I think that would be preferable. Elliot/Matt may have other ideas though?

fresto32:
We've made some adjustments to `setup.py`:
- subclassed `sdist` so we can download all binaries into `pact/bin` when running `python3 setup.py sdist`, so our users don't need to fetch the binaries if they install the dist artifact
- Kept the `--bin-path` for local development purposes
- Still allow the downloading of artifiacts for local development purposes

Please let me know if there's anything else we can do

mefellows:
Awesome, thanks so much! I'll leave it to the maintainers of pact-python to review.

setup.py
```diff
@@ -48,25 +70,66 @@ class PactPythonInstallCommand(install):
...
+
+    path = os.path.join(download_bin_path, binary['filename'])
+
+    if os.path.isfile(path) is True:
+        extract_ruby_app_binary(download_bin_path, package_bin_path, binary['filename'])
+    else:
+        if download_bin_path is not None:
+            if os.path.isfile(path) is not True:
+                raise RuntimeError('Could not find {} binary.'.format(path))
+            extract_ruby_app_binary(download_bin_path, package_bin_path, binary['filename'])
```
elliottmurray, line 119: Is this line reachable?

elliottmurray:
Sorry for being awol. This is looking pretty good. Thank you for the awesome contribution! Had one minor comment but I'm going to merge anyway.

fresto32:
Thanks Elliot! 🥳 Feels great to contribute to Pact.

setup.py
fresto32: I think this code will be hit if the user supplies a `bin-path` (i.e. `download_bin_path`) CLI argument and it contains the necessary standalone binaries

m-aciek: @fresto32 please note that in line 108 we check the opposite condition, so therefore `download_bin_path is not None` will never be false. I think the current behavior is broken, when I try to install, the binary is not downloaded currently. It worked for me after I removed the if and left:
```
download_ruby_app_binary(download_bin_path, binary['filename'], binary['suffix'])
extract_ruby_app_binary(download_bin_path, package_bin_path, binary['filename'])
```

PR #265, opened 2021-09-21
feat: ffi verifier cli

mikegeeves:
Here's where the code is up to, I've merged in recent changes which have gone into master (the updates to setup.py to handle downloading the binaries for example, so the rust ones can then follow the same approach). The CLI is working with the auto-generated click arguments and looks like it works ok, but aware there are concerns with this approach, so pushing over where this is currently up to to figure out what to do next.

Note: this is using a not yet released FFI version, so I was running with a built 0.0.3 FFI from pact-reference


mikegeeves:
Can't see what snyk is complaining about, is it possible to get access or limited number of users etc?

mefellows:
Hi @mikegeeves, sorry I missed this - have enabled notifications on this repo.

Do you need a second pair of eyes or anything over this?

JP-Ellis:
I'm going to close this PR as I don't think it is needed anymore with the release of Pact Python v3.

Thanks for all the amazing work though!

PR #266, opened 2021-09-21
feat: enable pending default true

mikegeeves:


bethesque:
Just pause this for a bit - we're having some disagreement about whether or not this is a good idea. Will comment back as soon as possible.

bethesque:
Thanks for getting on to this so quickly @mikegeeves  Please see note on the issue https://github.com/pact-foundation/pact-python/issues/264

elliottmurray:
> Thanks for getting on to this so quickly @mikegeeves Please see note on the issue #264

So @bethesque are you happy if I bump to 2.0.0? I can put current main on a versioned branch then.

bethesque:
Let's leave this for now, until we have a cohesive strategy. I'm also thinking of renaming it to ["smart_fail"](https://github.com/pact-foundation/pact_broker/discussions/507), as "pending pacts" causes so much confusion, and that would be a better time to bump the major version.

AbrahamGoRa8:
Hi @bethesque, jumping again about this problem. 

As currently, the default behaviour in the pact-broker is considering failures pacts as pending we don't have any way with pact-python to set up this to False. The current implementation in pact-python doesn't include the feature to disable this in the pact broker.

I saw in your PR, you were including the option for --no-enable-pending, without this option in the pact-python library. Currently, there is no way to disable the pending feature.

Should we add this feature (add --no-enable-pending) without changing the default value as False at least?

mefellows:
I thought we modified the broker's default configuration to disable that? I could be misremembring.

But in either case, an explicit flag here to disable it would be good if it doesn't exist already. Would you be up for creating a PR for this @AbrahamGoRa8 ?

AbrahamGoRa8:
> I thought we modified the broker's default configuration to disable that? I could be misremembring.
> 
> But in either case, an explicit flag here to disable it would be good if it doesn't exist already. Would you be up for creating a PR for this @AbrahamGoRa8 ?

Could you give me permissions to create the branch and opening the PR? I have got this error
ERROR: Permission to pact-foundation/pact-python.git denied to AbrahamGoRa8.

mikegeeves:
@AbrahamGoRa8 you'll need to fork the repo from pact-foundation, then submit the PR from there

bethesque:
I changed the default from false to true in the broker, not realising that the clients weren't sending a value through at all when it was false. Tim has convinced me that we should leave the default as false, but I'm not going to put out yet another release of the broker with the default changed. That's just going to screw too many people up. So yes, we definitely need a way to turn it off from each client.

PR #269, opened 2021-10-01
chore(ruby): update ruby standalone

m-aciek:
This should enable PACT_DISABLE_SSL_VERIFICATION=true for Pact broker verification.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ from distutils.command.sdist import sdist as sdist_orig
 
 
 IS_64 = sys.maxsize > 2 ** 32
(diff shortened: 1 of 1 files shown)


mefellows:
FYI I've released this as [v1.4.4](https://pypi.org/project/pact-python/1.4.4/).

mefellows:
`v1.4.5` should have the proper upstream fix - can you please check if this resolves the issue? You should no longer need to set environment variables manually.

m-aciek:
Confirming this resolves the issue. Thank you!

Issue #270, opened 2021-10-08
Examples: remove misleading use of PACT_MOCK variables in examples


mefellows:
The fastapi example (and perhaps the others) are a little misleading to newbies.

The following [lines](https://github.com/pact-foundation/pact-python/blob/master/examples/fastapi_e2e/tests/provider/test_provider.py#L23-L25) would suggest that there is pact mock on the provider side, when in reality this is the address of the locally running FastAPI provider:

```python
PACT_MOCK_HOST = 'localhost'
PACT_MOCK_PORT = 8000
PACT_URL = "http://{}:{}".format(PACT_MOCK_HOST, PACT_MOCK_PORT)
```

I'd suggest changing these to `PROVIDER_HOST`, `PROVIDER_PORT` and `PROVIDER_URL` respectively, perhaps with a comment explaining these map to the default host/port combination of the provider.

See this thread for context: https://pact-foundation.slack.com/archives/C9VECUP6E/p1633686950075000

mikegeeves:
I think this one should be ok to close off now, note the file is now at: https://github.com/pact-foundation/pact-python/blob/master/examples/fastapi_provider/tests/provider/test_provider.py

mefellows:
I this is good now.

PR #271, opened 2021-10-08
test(examples): tidy build and flask/fastapi examples, improve consistency

mikegeeves:
re #270


mikegeeves:
> This is looking really awesome btw. I'm trying to get it to run locally to help get it over the lone but having some issues with install to do with ruby binaries. I think an earlier commit added some options and has some unintended consequences I'm trying to unpick.

Thanks :)
It wasn't running the FastAPI examples before from the Makefile, and the scripts didn't quite work right so it might be from that, still very much WIP/Draft!
The consumer and Flask examples are pretty much there, FastAPI and messager I still need to look at, so half way there.
In the Flask example it's running flask from the run_pytest.sh script, but FastAPI is also running it as a fixture, need to figure out where to go with that one, I think probably the fixture.
Anyway, making progress

mikegeeves:
> So I originally set it up so there were a few different examples. The fixture one I added to show how to run with pytest. Also there is an issue that was beyond my python of getting Flask to run as a sub process to do with tickle (they made some changes that effectively broke it and the script was the only way I could get it to work)

I think that's ok, it's helpful to show different approaches, but can be confusing without the "why" being clear.
I'm going for more of a central examples README, rather than the individual ones per folder, to be able to compare/contrast them there, what do you think? Started but in progress

elliottmurray:
Just checking @mikegeeves I can close this as we've merged it?

mikegeeves:
Think it's fine for now to close yes, there will be more later but that can go separately

PR #276, opened 2021-11-24
docs(examples a->b->c): starting point of implementing Pact tests where there is a chain of services

mikegeeves:


mefellows:
Wow, this is a big one. I'll pop it on the TODO - but thank you for doing this 🙏

examples/abc/provider_b_hub/tests/provider/test_provider_b_provider.py
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+
+def test_provider_b_hub_against_broker(server, broker_opts, **kwargs):
+    verifier = Verifier(provider=PROVIDER_NAME, provider_base_url=PROVIDER_URL)
+
+    success, logs = verifier.verify_with_broker(
+        **broker_opts,
+        verbose=True,
+        provider_states_setup_url=f"{PROVIDER_URL}/_pact/provider_states",
```
joshua-badger, line 44: I'm trying to understand `provider_states_setup_url`, can you explain what it does or point me to documentation about it?

mikegeeves: I didn't manage to find as good documentation as I would like for that when I was trying to get it working, descriptions and examples in some languages but it probably needs more work. https://docs.pact.io/implementation_guides/python/readme#provider-states

The TLDR is you create an extra endpoint (created here: examples/abc/provider_b_hub/tests/pact_provider_b.py). When the verifier performs the tests, for each test it FIRST does a POST to that endpoint with {"state": "some books exist"} etc. You then define whatever you need to there to get the service into an appropriate state, so when the ACTUAL request is made, it can behave as desired.

I like how this is setup with FastAPI, to be able to only have this extra route for the purposes of testing, it gives a nice way to not have some super admin drop all tables and setup endpoint accidentally left open!

joshua-badger: Revisiting this, I think I can do something similar in Django. This is intriguing. Thanks!

YOU54F:
@mikegeeves nice work fella, will have a review this week

mikegeeves:
There's a long thread somewhere on this, I think it needs some re-work to shuffle around where Pact is wrapping around. Currently I'm mocking the http call, but from a long discussion with Tim it sounds like it would be better to move this up so it's wrapping a method around it rather than that low level. How to mock the http call like that is nice to know how to [fun figuring out getting that working :D] but I think higher up makes sense to then have a nicer pattern for the next step of async rather than HTTP where the pattern can be continued rather than completely different

JP-Ellis:
While this can't be merged as is, I have created #1274 to track the addition of a new example which can re-use some of the code you have here.

JP-Ellis:
Closing this as it has been re-implemented by #1528 using version 3.

Issue #241, opened 2021-07-02
Support bytes as response type


stefanello8:
I have a Consumer that expects bytes as response. Something like this:
`.will_respond_with(200, body=bytes([0x13, 0x00, 0x00, 0x00, 0x08, 0x00]))`

But it seems like bytes are not supported as response type of an interaction, since I'm getting a value error from pact (_ValueError: Unknown type: <class 'bytes'>_).

Is it possible to support bytes as well or is there any workaround for this kind of scenario?

joshua-badger:
I also have a situation this would be helpful for. I'd love to see this happen.

joshua-badger:
This issue has not actually been solved yet, the fix applied was naive in a few assumptions.

PR #281, opened 2022-01-03
feat(matcher): Allow bytes type in from_term function

joshua-badger:
Closes https://github.com/pact-foundation/pact-python/issues/241, wherein using bytes is not a valid option for either request or response content.

### DIFF
--- a/pact/matchers.py
+++ b/pact/matchers.py
@@ -184,7 +184,7 @@ def from_term(term):
     Parse the provided term into the JSON for the mock service.
 
     :param term: The term to be parsed.
-    :type term: None, list, dict, int, float, str, unicode, Matcher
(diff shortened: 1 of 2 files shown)


elliottmurray:
And this works for you? Just checking as I haven't run it locally

joshua-badger:
I am unable to run `make test` due to weirdness on OSX, but I can run the tests individually from Pycharm

elliottmurray:
More meant you were comfortable it is testing what you think it is. It looks like it.

joshua-badger:
Oh, yes. The traceback I ran into led me to that place, and we just need the return value to be what it is.

joshua-badger:
Thanks @elliottmurray !

elliottmurray:
No thank you!
 
At some point I'll fix my local env and trigger a release.

PR #284, opened 2022-01-27
feat: No include pending

AbrahamGoRa8:
* feat: Including new option to disable pending pacts

### DIFF
--- a/pact/__version__.py
+++ b/pact/__version__.py
@@ -1,3 +1,3 @@
 """Pact version info."""
 
-__version__ = '1.4.6'
+__version__ = '1.5.0'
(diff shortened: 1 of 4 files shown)


elliottmurray:
@AbrahamGoRa8 just for my sanity does test_all_url_options test the positive case (--enable-pending) as your test changes assert the negative?

bethesque (approved):

AbrahamGoRa8:
> @AbrahamGoRa8 just for my sanity does test_all_url_options test the positive case (--enable-pending) as your test changes assert the negative?

yeah, `test_all_url_options` is testing the positive case. The default behaviour will be `--no-enable-pending` as it was defined in the method statements so I had to add this new argument to all of the tests where the test was not sending the positive case but the mentioned one, it was sending the positive case so it's still testing that.

PR #289, opened 2022-02-14
feat: message_pact -> with_metadata() updated to accept term

sunsathish88:


mefellows:
It's worth checking if this actually works. I don't believe matchers currently work in the core Ruby application that does the matching rules. I don't think that should block this PR coming through, but worth knowing for sure.

sunsathish88:
> It's worth checking if this actually works. I don't believe matchers currently work in the core Ruby application that does the matching rules. I don't think that should block this PR coming through, but worth knowing for sure.

I have actually updated the meta_data under message_pact with from_term() to transform to JSON. It is not the matchers, the title must have added confusion. I did test this out locally and it works.

sunsathish88:
Is there a timeline when this PR can be merged and released (if no changes required) ? Appreciate the response!

elliottmurray:
It took me a bit to get my head around what you're doing but make sense now. I would like a test added if that's ok? I'll then approve and release. 

If you go to test_message_pact and one of the tests alter to have a term as a value of with_metadata e.g.:

`.with_metadata({'contentType': 'application/json',
                            'source': 'legacy_api',
                            'some-header': Term('\d+-\d+-\d+T\d+:\d+:\d+', '2022-02-15T20:16:01')})`

You don't have to use a complex date regex so feel free to simplify. Then you should be able to  make a couple of small changes like:
`
        self.assertTrue({'contentType': 'application/json', 'source': 'legacy_api'}.items()
                        <= target._messages[0]['metaData'].items())
        self.assertTrue(any([True for k, v in target._messages[0]['metaData']['some-header'].items() if v == 'Pact::Term']))
`

This now tests the previous items still exist and the additional element is added and is a Pact Term element. I'll accept that your new header item contains a dict with a key with 'Pact::Term' as good enough though open to improvements.

sunsathish88:
@elliottmurray Thanks! Yeah sure I can get that added. Do you want me to add a separate test case or modify an existing one to include the pact term ?

elliottmurray:
> @elliottmurray Thanks! Yeah sure I can get that added. Do you want me to add a separate test case or modify an existing one to include the pact term ?

When I looked it it I think modifying one of the message pact tests would be enough. It's a slight mod to an existing 'functiuonal style' unit test.

sunsathish88:
Thanks @elliottmurray ! I have updated the tests and ran it locally to verify. Please let me know if this looks good.

elliottmurray (approved):

elliottmurray:
@sunsathish88 a flake issue - might need to format a couple of lines (just run flake on the cmd line when done and it will tell you if it will pass or not)

PR #295, opened 2022-03-24
fix: call kill instead of terminate in message_provider

### LINKED ISSUE #294: _wait_for_server_stop cannot stop process and throw AssertionError
There is a docker-compose configuration with django application.
When inside docker I run `pytest` command to execute `pact` tests is fails on `_wait_for_server_stop` call. Looks like that process cannot be stopped for some reason.

The problem is that it's happening only for the first time when I build containers. All the following execution of `pytest` completes successfully.

I'm trying to figre out the reason, but without success.

This is a stacktrace:
```
__________ TestProvider.test_something __________

self = <pact.message_provider.MessageProvider object at 0x7fc61e720430>

    def _wait_for_server_stop(self):
        """Wait for server to finish, or raise exception after timeout."""
        retry = 20
        while True:
            self._process.terminate()
            time.sleep(0.1)
            try:
>               assert not self._process.is_alive()
E               AssertionError

/usr/local/lib/python3.10/site-packages/pact/message_provider.py:92: AssertionError

During handling of the above exception, another exception occurred:

    def test_something():
       provider = MessageProvider(
            message_providers={
                'something triggered': message_provider
            },
            provider='api',
            consumer='consumer',
            pact_dir='./_pacts',
        )

>       with provider:

application/tests/services/test_pact.py:100:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/local/lib/python3.10/site-packages/pact/message_provider.py:156: in __exit__
    self._stop_proxy()
/usr/local/lib/python3.10/site-packages/pact/message_provider.py:108: in _stop_proxy
    self._wait_for_server_stop()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pact.message_provider.MessageProvider object at 0x7fc61e720430>

    def _wait_for_server_stop(self):
        """Wait for server to finish, or raise exception after timeout."""
        retry = 20
        while True:
            self._process.terminate()
            time.sleep(0.1)
            try:
                assert not self._process.is_alive()
                return 0
            except AssertionError:
                if retry == 0:
>                   raise RuntimeError("Process timed out")
E                   RuntimeError: Process timed out

/usr/local/lib/python3.10/site-packages/pact/message_provider.py:96: RuntimeError
----------------------------- Captured stdout call -----------------------------
INFO: Reading pact at ./_pacts/api-consumer.json


Verifying a pact between api and consumer
  Given something triggered
    message-name
WARN: Skipping set up for provider state 'something triggered' for consumer 'consumer' as there is no --provider-states-setup-url specified.
      has matching content

1 interaction, 0 failures
=========================== short test summary info ============================
FAILED application/tests/services/test_pact.py::TestProvider::test_something
```

sentyaev:
Closes #294 

When run pact tests inside fresh docker container `_wait_for_server_stop` function throwing an `AssertionError` because process is not stopped. Looks like `SIGTERM` signal is captured by another process. `process.kill()` which sends `SIGKILL` works.

### DIFF
--- a/pact/message_provider.py
+++ b/pact/message_provider.py
@@ -86,7 +86,7 @@ class MessageProvider(object):
         """Wait for server to finish, or raise exception after timeout."""
         retry = 20
         while True:
-            self._process.terminate()
+            self._process.kill()
             time.sleep(0.1)
             try:
                 assert not self._process.is_alive()


YOU54F:
Thanks for the PR @sentyaev,

Should we consider the comment here https://github.com/pact-foundation/pact-python/issues/294#issuecomment-1076595755

sentyaev:
This PR not needed. Issue solved by removing `freeze_time` decorator from the test function.

PR #296, opened 2022-04-10
ci: added poetry lock file

elliottmurray:


sergeyklay:
I'm going to close this pull request since the original author seems to have lost interest in it over a year ago, and I don't want a collection of half-finished work to accumulate here. I'm open to add support for anything/everything at some point that have a well-justified reason, which should be described in an issue. However, this pull request was opened without providing a clear rationale for the changes, and it's not apparent to me what problem it is intended to solve or the priority of these changes.

If anyone believe that these changes are still relevant and necessary, please feel free to open a new issue describing the problem you're trying to solve and the proposed solution. This will help me better understand the context and prioritize the work accordingly.

Thank you for your understanding.

PR #297, opened 2022-06-18
build: Correct download logic when installing. Add a helper target to…

mikegeeves:
… setup a pyenv via make


mikegeeves:
Looking at the message examples, but had some blockers of getting a dev environment going again :slightly_smiling_face: I had this previously but worked around by just putting the .tar.gz file where it expected to be - the logic had download_bin_path always being set so the route to go get could never happen

PR #301, opened 2022-08-23
chore(test): fix consumer message test

B3nnyL:
Current consumer message example publishes pact with the same version number twice, update the fixture to ensure pact only publish once.


elliottmurray (approved):
Thanks - was going to solve this a different way but this works!

PR #319, opened 2022-11-23
feat: update standalone to 1.91.0

bethesque:


mefellows:
Did we remove support for linux x86 recently?

See https://github.com/pact-foundation/pact-python/pull/319 (which fixes it)

YOU54F:
Yes! but its back :) in 2.0.0!

PR #320, opened 2022-11-23
Remove support for x86 linux

mefellows:
Fixes failing build here: https://github.com/pact-foundation/pact-python/pull/319

See https://github.com/pact-foundation/pact-ruby-standalone/pull/75

Travelling ruby has deprecated support for linux-x86 (x86_64 is still supported), so that's been removed from the created packages. 

This PR essentially drops support for linux x86. 

This _probably_ ought to be a major version bump, but I'll leave that to the maintainer to decide. I suspect not many people use this architecture.


YOU54F:
No longer needed with #342 as this restores x86 packages in travelling ruby.

I am not sure how many users we having using 32 bit, and whether it would be supporting in the Pact FFI officially (I have some unofficial builds cooked up)

PR #322, opened 2022-11-26
feat: Enhance provider states for pact-message

nsfrias:
* feat: allow multiple provider states to be defined in a pact

* feat: allow additional parameters to be passed in the provider state

Signed-off-by: Nuno Frias <nsfrias@gmail.com>

### DIFF
--- a/pact/message_pact.py
+++ b/pact/message_pact.py
@@ -98,22 +98,31 @@ class MessagePact(Broker):
         self._messages = []
         self._message_process = None
 
-    def given(self, provider_states):
+    def given(self, name, params=None):
         """
         Define the provider state for this pact.
 
         When the provider verifies this contract, they will use this field to
         setup pre-defined data that will satisfy the response expectations.
 
-        :param provider_state: The short sentence that is unique to describe
-            the provider state for this contract.
-        :type provider_state: basestring
+        :param name: The short sentence that is unique to describe the provider
+        state for this contract.
+        :type name: basestring
+        :param params: Additional arguments necessary to set the porovider state
+        :type params: dict
         :rtype: Pact
         """
         self._insert_message_if_complete()
 
-        state = [{"name": "{}".format(provider_states)}]
-        self._messages[0]['providerStates'] = state
+        provider_state = {'name': "{}".format(name)}
+        if params:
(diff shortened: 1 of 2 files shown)


mefellows:
Thank you! I'm surprised this works, because Ruby doesn't support V3 contracts (properly) but I'm assuming it does work otherwise you wouldn't have submitted the PR.

I think it's OK to support this, assuming the code quality/review passes.

pact/message_pact.py
```diff
@@ -98,22 +98,31 @@ def __init__(
...
+        :type params: dict
         :rtype: Pact
         """
         self._insert_message_if_complete()
 
-        state = [{"name": "{}".format(provider_states)}]
-        self._messages[0]['providerStates'] = state
+        provider_state = {'name': "{}".format(name)}
+        if params:
+            provider_state['params'] = from_term(params)
```
mikegeeves, line 119: Is using from_term here meaningful? I'm not sure if matchers make sense for the params, they need to be fixed don't they?

tests/test_message_pact.py
```diff
@@ -82,6 +82,60 @@ def test_definition_sparse(self):
 
         self.assertTrue(target._messages[0]['metaData']['some-header'], 'Pact::Term')
 
+    def test_definition_multiple_provider_states(self):
+        target = MessagePact(self.consumer, self.provider)
+        (
+            target
+            .given('there is an alligator named John',
+                params={'color':'green', 'weight_kg': 130, 'lenght_m': 1.95}
```
mikegeeves, line 90: Typo, should be length rather than lenght (repeated in a few places)

mikegeeves:
It looks fine as far as I can see [haven't tried testing it manually though]

TIL: using <= for testing for subsets, had not seen that before thanks! :D

nsfrias:
@mikegeeves All comments addressed. Btw the `<=` was copied from another existent test. It was new to me too.

nsfrias:
@mikegeeves is it possible to merge, assuming you're happy with my fixes.

mikegeeves (approved):
Looks fine to me :)

elliottmurray (approved):
Based on Mikes comments

PR #329, opened 2023-03-05
fix: replace yanked version of wheel with a more recent one

sergeyklay:
Reason for being yanked: Circular dependency with setuptools on --no-binary=:all:


sergeyklay:
Close in favour of https://github.com/pact-foundation/pact-python/pull/331

PR #330, opened 2023-03-05
fix: remove dead code

sergeyklay:
- `missing_files` was set twice, the second time is not used
- `provider_app_version` was set twice with the same value
- removed unnecessary brackets
- fix link for GitHub badge


mike-geeves-oxb (approved):
LGTM

Standardising READMEs would be nice to try and do, all of the different projects look a little different, but that's for another PR :)

mikegeeves (approved):
Let's try from the right account 🤦 

LGTM

Standardising READMEs would be nice to try and do, all of the different projects look a little different, but that's for another PR :)

PR #311, opened 2022-10-03
add support for Python 3.11

gruebel:
there was one deprecation warning when I ran `tox -e py311` it was related to an old `coverage` version, so updated it to a newer one, where it was fixed 🙂

PR #331, opened 2023-03-05
ci: add python 3.11 to test matrix

sergeyklay:
- add python 3.11 to test matrix and setup.py
- describe classifiers and python version for pypi package
- replace yanked version of wheel with a more recent one
- use compatible dependency versions for Python 3.6

closes #311
closes #329

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4,29 +4,50 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


mikegeeves (approved):
I was wondering if it's helpful to leave 3.6 in vs just remove the extra logic, but it's probably helpful to have a way of having a "deprecated" version to just roll as it goes.
3.7 expires in about 2.5 months so we can remove 3.6 and deprecate 3.7 accordingly then :)

LGTM!

PR #333, opened 2023-03-12
feat: add matchers for ISO 8601 date format

sergeyklay:
This introduces `pact.Format.iso_8601_datetime()` method to match a string for a full ISO 8601 Date.

This method does not do any sort of date validation, only checks if the string is according to the ISO 8601 spec.

It differs from `pact.Format.timestamp`, `pact.Format.date` and `pact.Format.time` implementations in that it is more stringent and tests the string for exact match to the ISO 8601 dates format.

Without `with_ms` parameter will match string containing ISO 8601 formatted dates as stated bellow:

* 2016-12-15T20:16:01
* 2010-05-01T01:14:31.876
* 2016-05-24T15:54:14.00000Z
* 1994-11-05T08:15:30-05:00
* 2002-01-31T23:00:00.1234-02:00
* 1991-02-20T06:35:26.079043+00:00

Otherwise, ONLY dates with milliseconds will match the pattern:

* 2010-05-01T01:14:31.876
* 2016-05-24T15:54:14.00000Z
* 2002-01-31T23:00:00.1234-02:00
* 1991-02-20T06:35:26.079043+00:00

This change aims to bring the capabilities of the python library into alignment with pact-foundation/docs.pact.io#88, since the existing functionality is a bit liberal and allows tests to pass even in cases where the dates do not conform to the ISO 8601 spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,23 +257,25 @@ Often times, you find yourself having to re-write regular expressions for common
 ```python
 from pact import Format
 Format().integer  # Matches if the value is an integer
-Format().ip_address  # Matches if the value is a ip address
+Format().ip_address  # Matches if the value is an ip address
 ```
 
 We've created a number of them for you to save you the time:
 
-| matcher          | description                                                                                     |
-|-----------------|-------------------------------------------------------------------------------------------------|
-| `identifier`  | Match an ID (e.g. 42)                                                                           |
-| `integer`     | Match all numbers that are integers (both ints and longs)                                       |
-| `decimal`     | Match all real numbers (floating point and decimal)                                             |
-| `hexadecimal`    | Match all hexadecimal encoded strings                                                           |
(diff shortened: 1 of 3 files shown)


mefellows (approved):
Looks good! The only comment I have is that other languages let you specify the example value, whereas in this matcher it's hard coded.

It's consistent with the other matchers, so I think it's OK, but wanted to share that feedback.

sergeyklay:
@mefellows I'll definitely fix this, thank you for your comment. However, I would like to address it in a separate pull request. The issue is that there is currently hardcoded logic in all matchers, and I'd prefer to make a single, unified change to address this.

mefellows:
Absolutely! Let's not hold up this change for that.

YOU54F (approved):

YOU54F:
Nice one, we've got a flakey build, but you get a mergey merge ( i just got one merging a docco change :( )

YOU54F:
also thanks for updating the readme and adding a test

PR #334, opened 2023-03-12
feat: provide ability to verify with a specific pact

### LINKED ISSUE #332: No way to verify with a specific pact
I'm using `pact-python` version 1.7.0.
There doesn't seem to be a way to verify with a specific pact by passing in a pact URL.

I tried passing a `pact_url` kwarg into `MessageProvider.verify_with_broker` but this always gives me the error `pact.verify_wrapper.PactException: Pact urls or Pact broker required`.

sergeyklay:
This change adds the ability to verify pacts by passing their URI to verify. The URI of the every pact can be an HTTP URI or a local file path.

Fixes #332

### DIFF
--- a/pact/message_provider.py
+++ b/pact/message_provider.py
@@ -115,22 +115,37 @@ class MessageProvider(object):
         return_code, _ = verifier.verify_pacts(pact_files, verbose=False)
         assert (return_code == 0), f'Expected returned_code = 0, actual = {return_code}'
 
-    def verify_with_broker(self, enable_pending=False, include_wip_pacts_since=None, **kwargs):
+    def verify_with_broker(
+            self,
+            *pacts,
+            enable_pending=False,
+            include_wip_pacts_since=None,
+            **kwargs
+    ):
         """Use Broker to verify.
 
-        Args:
-            broker_username ([String]): broker username
-            broker_password ([String]): broker password
-            broker_url ([String]): url of broker
-            enable_pending ([Boolean])
-            include_wip_pacts_since ([String])
-            publish_version ([String])
-
+        :param pacts: List of pact to verify. Every pact in that list can be
+            an HTTP URI or a local file.
+        :type pacts: str
+        :param enable_pending: Allow pacts which are in pending state to be
+            verified without causing the overall task to fail.
+        :type enable_pending: bool
+        :param include_wip_pacts_since: Allow pacts that don't match given
+            consumer selectors (or tags) to  be verified, without causing the
+            overall task to fail
+        :type include_wip_pacts_since: str or None
         """
-        verifier = Verifier(provider=self.provider,
-                            provider_base_url=self._proxy_url())
+        verifier = Verifier(
+            provider=self.provider,
+            provider_base_url=self._proxy_url()
+        )
 
-        return_code, _ = verifier.verify_with_broker(enable_pending, include_wip_pacts_since, **kwargs)
+        return_code, _ = verifier.verify_with_broker(
+            *pacts,
+            enable_pending=enable_pending,
+            include_wip_pacts_since=include_wip_pacts_since,
+            **kwargs
+        )
 
         assert (return_code == 0), f'Expected returned_code = 0, actual = {return_code}'
 
@@ -147,7 +162,7 @@ class MessageProvider(object):
         Exit a Python context.
 
         Return False to cascade the exception in context manager's body.
-        Otherwise it will be supressed and the test will always pass.
+        Otherwise, it will be suppressed and the test will always pass.
         """
         if (exc_type, exc_val, exc_tb) != (None, None, None):
             if exc_type is not None:
--- a/pact/message_provider.py
+++ b/pact/message_provider.py
@@ -125,7 +125,7 @@ class MessageProvider(object):
         """Use Broker to verify.
 
         :param pacts: List of pact to verify. Every pact in that list can be
-            an HTTP URI or a local file.
+            an HTTP URI.
         :type pacts: str
         :param enable_pending: Allow pacts which are in pending state to be
             verified without causing the overall task to fail.
--- a/pact/verifier.py
+++ b/pact/verifier.py
@@ -28,10 +28,17 @@ class Verifier(object):
         return 'Verifier for {} with url {}'.format(self.provider, self.provider_base_url)
 
     def validate_publish(self, **kwargs):
-        """Validate publish has a version."""
-        if (kwargs.get('publish') is not None) and (kwargs.get('publish_version') is None):
-            # do something
-            raise Exception()
+        """Ensure the participant version is provided.
+
+        :raises Exception: When `publish_verification_results` is set,
+            but `publish_version` is not provided.
+        """
+        if not kwargs.get('publish_verification_results'):
+            return
+
+        if not kwargs.get('publish_version'):
+            raise Exception('To publish the results of verification to the '
+                            'broker version of the participant is mandatory')
 
     def verify_pacts(
             self,
@@ -70,6 +77,8 @@ class Verifier(object):
             consumer selectors (or tags) to  be verified, without causing the
             overall task to fail
         :type include_wip_pacts_since: str or None
+        :raises Exception: When `publish_verification_results` is set,
+            but `publish_version` is not provided.
         :return: Returns a tuple of two elements. The first indicates the
             status of the operation, so that 0 means success, and the second
             contains the logs of the operation.
@@ -142,7 +151,8 @@ class Verifier(object):
         :keyword str broker_token: Pact Broker bearer token.
         :keyword str broker_url: Base URL of the Pact Broker from which to
             retrieve the pacts.
-        :raises Exception: [ErrorDescription]
+        :raises Exception: When `publish_verification_results` is set,
+            but `publish_version` is not provided.
         :return: Returns a tuple of two elements. The first indicates the
             status of the operation, so that 0 means success, and the second
             contains the logs of the operation.
--- a/pact/verifier.py
+++ b/pact/verifier.py
@@ -42,7 +42,7 @@ class Verifier(object):
     ):
         """Verify pacts from the provider.
 
-        Example:
+        Usage:
 
         >>> from pact import Verifier
         >>>
@@ -106,7 +106,7 @@ class Verifier(object):
     ):
         """Use Broker to verify.
 
-        Example:
+        Usage:
 
         >>> from pact import Verifier
         >>>
--- a/pact/verifier.py
+++ b/pact/verifier.py
@@ -33,63 +33,149 @@ class Verifier(object):
             # do something
             raise Exception()
 
-    def verify_pacts(self, *pacts, enable_pending=False, include_wip_pacts_since=None, **kwargs):
-        """Verify our pacts from the provider.
+    def verify_pacts(
+            self,
+            *pacts,
+            enable_pending=False,
+            include_wip_pacts_since=None,
+            **kwargs
+    ):
+        """Verify pacts from the provider.
 
-        Returns:
-          success: True if no failures
-          logs: some tbd output of logs
+        Example:
 
+        >>> from pact import Verifier
+        >>>
+        >>> PROVIDER_URL = f'http://127.0.0.1:5001'
+        >>>
+        >>> verifier = Verifier(
+        ...   provider='AcmeAPI',
+        ...   provider_base_url=PROVIDER_URL
+        ... )
+        >>>
+        >>> result, _ = verifier.verify_pacts(
+        ...   'consumer-provider.json',
+        ...   enable_pending=True,
+        ... )
+        >>>
+        >>> assert result == 0
+
+        :param pacts: List of pact to verify. Every pact in that list can be
+            an HTTP URI or a local file.
+        :type pacts: str
+        :param enable_pending: Allow pacts which are in pending state to be
+            verified without causing the overall task to fail.
+        :type enable_pending: bool
+        :param include_wip_pacts_since: Allow pacts that don't match given
+            consumer selectors (or tags) to  be verified, without causing the
+            overall task to fail
+        :type include_wip_pacts_since: str or None
+        :return: Returns a tuple of two elements. The first indicates the
+            status of the operation, so that 0 means success, and the second
+            contains the logs of the operation.
+        :rtype: tuple
         """
         self.validate_publish(**kwargs)
 
-        missing_files = [path for path in pacts if not path_exists(path)]
-        if missing_files:
-            raise Exception("Missing pact files {}".format(missing_files))
+        missing_pacts = [path for path in pacts if not path_exists(path)]
+        if missing_pacts:
+            raise Exception("Missing pact files: {}".format(
+                ', '.join(missing_pacts))
+            )
 
-        pacts = expand_directories(pacts)
-
+        pacts = expand_directories(list(pacts))
         options = self.extract_params(**kwargs)
-        success, logs = VerifyWrapper().call_verify(*pacts,
-                                                    provider=self.provider,
-                                                    provider_base_url=self.provider_base_url,
-                                                    enable_pending=enable_pending,
-                                                    include_wip_pacts_since=include_wip_pacts_since,
-                                                    **options)
 
-        return success, logs
+        return_code, logs = VerifyWrapper().call_verify(
+            *pacts,
+            provider=self.provider,
+            provider_base_url=self.provider_base_url,
+            enable_pending=enable_pending,
+            include_wip_pacts_since=include_wip_pacts_since,
+            **options
+        )
 
-    def verify_with_broker(self, enable_pending=False, include_wip_pacts_since=None, **kwargs):
+        return return_code, logs
+
+    def verify_with_broker(
+            self,
+            *pacts,
+            enable_pending=False,
+            include_wip_pacts_since=None,
+            **kwargs
+    ):
         """Use Broker to verify.
 
-        Args:
-            broker_username ([String]): broker username
-            broker_password ([String]): broker password
-            broker_url ([String]): url of broker
-            enable_pending ([Boolean])
-            include_wip_pacts_since ([String])
-            publish_version ([String])
+        Example:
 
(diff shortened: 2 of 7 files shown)


pact/message_provider.py
```diff
@@ -115,22 +115,37 @@ def verify(self):
         return_code, _ = verifier.verify_pacts(pact_files, verbose=False)
         assert (return_code == 0), f'Expected returned_code = 0, actual = {return_code}'
 
-    def verify_with_broker(self, enable_pending=False, include_wip_pacts_since=None, **kwargs):
+    def verify_with_broker(
+            self,
+            *pacts,
```
mefellows, line 120: Is this a breaking change, because it's a new required positional argument?

mefellows (commented):
I'll leave the Python code to the maintainer, thanks for raising this! It looks like you've just exposed `pacts` and plumbed the argument to the underlying verification process that already supports it.

My only comment is around whether or not this constitutes a breaking semver change in Python.

pact/message_provider.py
sergeyklay: It still works in the usual way, e.g.
```
verify_with_broker(
  enable_pending=False,
  include_wip_pacts_since=None,
  # other kwargs
)
``

Comment (sergeyklay) [reply pact/message_provider.py:120]: Just to demonstrate the idea:

```py
% python
Python 3.11.0 (main, Dec  3 2022, 14:47:23) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def test(*args, **kwargs):
...   print(args, kwargs)
...
>>> test()
() {}
>>>
>>> test('a', 'b', 'c')
('a', 'b', 'c') {}
>>>
>>> test('a', 'b', 'c', foo='bar', baz='buz')
('a', 'b', 'c') {'foo': 'bar', 'baz': 'buz'}
>>>
>>> test(foo='bar', baz='buz')
() {'foo': 'bar', 'baz': 'buz'}
>>>
```

mefellows: Awesome, thanks. TIL

pact/verifier.py
```diff
@@ -28,68 +28,160 @@ def __str__(self):
...
+        >>>
+        >>> result, _ = verifier.verify_pacts(
+        ...   'consumer-provider.json',  # Optional
+        ...   enable_pending=True,
+        ... )
+        >>>
+        >>> assert result == 0
+
+        :param pacts: Optional list of pacts to verify. Every pact in that list
+            can be an HTTP URI or a local file.
```
mikegeeves, line 69:
```suggestion
            can be an HTTP(S) URI, a local file, or a local directory.
```

pact/verifier.py
```diff
@@ -28,68 +28,160 @@ def __str__(self):
...
+        ... )
+        >>>
+        >>> result, _ = verifier.verify_pacts(
+        ...   'consumer-provider.json',  # Optional
+        ...   enable_pending=True,
+        ... )
+        >>>
+        >>> assert result == 0
+
+        :param pacts: Optional list of pacts to verify. Every pact in that list
```
mikegeeves, line 68: I don't think it is actually optional, because if you have no pacts defined here, and there is no broker (if using with a broker you would call verify_with_broker instead I think) it will fail in `verify_wrapper._validate_input`

pact/verifier.py
```diff
@@ -28,68 +28,160 @@ def __str__(self):
...
-                                                    provider_base_url=self.provider_base_url,
-                                                    enable_pending=enable_pending,
-                                                    include_wip_pacts_since=include_wip_pacts_since,
-                                                    **options)
-
-        return success, logs
 
-    def verify_with_broker(self, enable_pending=False, include_wip_pacts_since=None, **kwargs):
+        return_code, logs = VerifyWrapper().call_verify(
+            *pacts,
```
mikegeeves, line 97: In which case, this one (verify_with_broker) doesn't make sense to have *pacts, because we should use the broker instead.
I don't think it makes sense to try and do both? (verify with files as well as a broker)
(that question probably needs input from someone with wider Pact knowledge)

pact/verifier.py
```diff
@@ -28,68 +28,160 @@ def __str__(self):
...
+        ...   enable_pending=True,
+        ...   broker_username='pactbroker',
+        ...   broker_password='pactbroker',
+        ...   broker_url='http://localhost',
+        ... )
+        >>>
+        >>> assert result == 0
+
+        :param pacts: Optional list of pacts to verify. Every pact in that list
+            can be an HTTP URI or a local file.
```
mikegeeves, line 136:
```suggestion
            can be an HTTP(S) URI, a local file, or a local directory.
```

pact/verifier.py
```diff
@@ -28,68 +28,160 @@ def __str__(self):
...
         options = {
-            'broker_password': broker_password,
-            'broker_username': broker_username,
-            'broker_token': broker_token,
-            'broker_url': broker_url
+            'broker_password': kwargs.get('broker_password', None),
+            'broker_username': kwargs.get('broker_username', None),
+            'broker_token': kwargs.get('broker_token', None),
+            'broker_url': kwargs.get('broker_url', None),
         }
```
mikegeeves, line 164: I think all these args could go into `extract_params` along with the others and avoid needing the two steps for options/update options, both `verify` and `verify_with_broker` then go onto do the same approach to pull out the options and use if they are provided or not

pact/message_provider.py
```diff
@@ -115,22 +115,37 @@ def verify(self):
...
+        :param pacts: List of pact to verify. Every pact in that list can be
+            an HTTP URI or a local file.
+        :type pacts: str
+        :param enable_pending: Allow pacts which are in pending state to be
+            verified without causing the overall task to fail.
+        :type enable_pending: bool
+        :param include_wip_pacts_since: Allow pacts that don't match given
+            consumer selectors (or tags) to  be verified, without causing the
+            overall task to fail
+        :type include_wip_pacts_since: str or None
```
mikegeeves, line 136: I don't think we have any formal chosen docstring style, we should discuss this :)

I tend to end up using Google or reST, generally prefer reST, but I think type hints in the code rather than :type gives easier to read/maintain...but let's chat!

pact/verifier.py
```diff
@@ -28,68 +28,160 @@ def __str__(self):
...
 
+        :raises Exception: When `publish_verification_results` is set,
+            but `publish_version` is not provided.
+        """
+        if not kwargs.get('publish_verification_results'):
+            return
+
+        if not kwargs.get('publish_version'):
+            raise Exception('To publish the results of verification to the '
+                            'broker version of the participant is mandatory')
```
mikegeeves, line 41: What about something like
```suggestion
    raise Exception('To publish the results of verification to the broker, '
                    'attribute publish_version (containing the provider '
                    'version) must be set')
```
I find it helpful if the error tells me what I need to set, so at least if I'm not sure where/how I have something to search for 🙂

pact/verifier.py
```diff
@@ -28,68 +28,160 @@ def __str__(self):
...
+        :keyword str broker_username: Pact Broker basic auth username.
+        :keyword str broker_password: Pact Broker basic auth password.
+        :keyword str broker_token: Pact Broker bearer token.
+        :keyword str broker_url: Base URL of the Pact Broker from which to
+            retrieve the pacts.
+        :raises Exception: When `publish_verification_results` is set,
+            but `publish_version` is not provided.
+        :return: Returns a tuple of two elements. The first indicates the
+            status of the operation, so that 0 means success, and the second
+            contains the logs of the operation.
```
mikegeeves, line 154: Maybe a bit shorter:
```suggestion
        :return: Returns a tuple of the overall result (where 0 indicates success), and logs produced.
```

pact/verifier.py
sergeyklay: No it isn't. Right now you can call this function without this parameter and it will work correctly. Besides, my edits didn't change anything in this regard

pact/verifier.py
sergeyklay: I don't want to change this API somehow in the current Pull Request. I've put a bit of cleanup here, it's a code cleanup and nothing more. Please, let's do such things in a separate Pull Request,  and here I don't even want to discuss this issue, because it's offtopic

pact/verifier.py
mikegeeves: What do you mean will work correctly?
If I call it without anything for pacts:
```
verifier = Verifier(
           provider='UserService',
           provider_base_url='http://localhost:5001'
)
result, _ = verifier.verify_pacts()
```

It will error:
```
    def _validate_input(self, pacts, **kwargs):
        if len(pacts) == 0 and not self._broker_present(**kwargs):
>           raise PactException('Pact urls or Pact broker required')
E           pact.verify_wrapper.PactException: Pact urls or Pact broker required

../pact/verify_wrapper.py:135: PactException
```

There must be at least one pact provided, it isn't `Optional`. I am talking about the description, it shouldn't be described as optional when it will not work without...

pact/message_provider.py
YOU54F: What is the current style in the project? 

Do either of you have a preference?

Both maintenance for maintainers and end user experience in IDE's is important.

YOU54F:
Just a couple of notes before I dig into a review/testing at some point in the next couple of weeks.

- It is possible to verify a pact locally by direct path, or directory , but you would not publish verification results to a broker in this instance, as the pact hasn't come from a broker
- It is possible to verify to a pact by URL, which has come from a broker
  - The user should be able to configure whether verification results are published in this instance
  - The options for `consumerVersionSelectors` / `enablePending` / `includingWipPactsSince` are not relevant for verifications by pact url
    - There are two verification tasks
      - One by consumer pact change, which takes a URL provided by webhook
      - One by provider commit, which uses `ConsumerVersionsSelectors` to retrieve pacts to verify
      
      
I'm not sure how much the ruby backend guards you, if you
- read a local file, and try to publish verification results
- provide both a `pactUrl` and `consumerVersionSelectors` / `enablePending` / `includingWipPactsSince`
  - Should it error, should it use both, Should it warn you about one, and take precedence (if so, pactUrl would seem sensible, as it is something that wouldn't exist every time, and we should be able to test if its a local file and therefore not publish results)
  
Some of these questions are wider than this PR alone, and raised lifecycle questions about both backends (Ruby and Rust) so I need to do some testing on the cores directly (probably via the CLI tools) and see what happens

pact/message_provider.py
sergeyklay: The default documentation style of reST in sphinx is supported by IDEs, used by the community and as far as I understand @mikegeeves is OK to forward with it too.

pact/verifier.py
sergeyklay: I see. I corrected it. Thank you for your patience.

requirements_dev.txt
```diff
@@ -15,7 +15,22 @@ pytest==7.1.3; python_version >= '3.7'
...
+# 4. docker/utils/decorators.py:_get()
+# 5. docker/api/client.py:get()
+# 6. requests/sessions.py:request()
+# 7. requests/sessions.py:send()
+# 8. requests/adapters.py:urlopen() <- calls _make_request() with 'chunked' param
+# 9. urllib3/connectionpool.py:_make_request() <- TypeError: request() got an unexpected keyword argument 'chunked'
+#
+# For more see:
+# https://github.com/pact-foundation/pact-python/actions/runs/4874690954/jobs/8696170180?pr=334
+urllib3>=1.26.12,<2
```
sergeyklay, line 33: This change isn't related to the current pull request. I discovered the crashing tests today. Probably because requests-2.30.0 (which uses urllib3 2.0 by default) is released. I'll be able to figure it out later.

sergeyklay: requests 2.30.0 version was yanked 15 min ago
https://github.com/psf/requests/issues/6439

pact/message_provider.py
```diff
@@ -115,22 +115,37 @@ def verify(self):
...
-        Args:
-            broker_username ([String]): broker username
-            broker_password ([String]): broker password
-            broker_url ([String]): url of broker
-            enable_pending ([Boolean])
-            include_wip_pacts_since ([String])
-            publish_version ([String])
-
+        :param pacts: List of pact to verify. Every pact in that list can be
+            an HTTP URI or a local file.
```
YOU54F, line 128: This should read, HTTP URI, and not a local file. There is no use case for reading a local file, and publishing the results to a pact broker, as the contents of the local pact, may not relate to the actual content from the broker.

```
List of pact to verify. Every pact in that list can be
            an HTTP URI
```

pact/message_provider.py
```diff
@@ -115,22 +115,37 @@ def verify(self):
...
-                            provider_base_url=self._proxy_url())
-
-        return_code, _ = verifier.verify_with_broker(enable_pending, include_wip_pacts_since, **kwargs)
+        verifier = Verifier(
+            provider=self.provider,
+            provider_base_url=self._proxy_url()
+        )
+
+        return_code, _ = verifier.verify_with_broker(
+            *pacts,
```
YOU54F, line 144: it should be one pact url only, the idea being that the pact broker would trigger an event an in turn a webhook, to the provider build with the pact url to verify (singular).

https://docs.pact.io/pact_broker/webhooks#events

there may be multiple webhooks fired, once for the main branch of the provider and once for any deployed or released versions.

We should be clear about this or guard against it being a non uri, and not a local file.

pact/verifier.py
YOU54F: The original request was via URL

https://docs.pact.io/pact_broker/webhooks#events

They didn't mention the file system here

YOU54F (changes requested):
We should only support one url for verification with a broker

>   To verify a pact at a known URL (eg. when a verification is triggered by a 'contract content changed'
  webhook), pass in the pact URL(s) as the first argument(s) to the command, and do NOT set any of the 
  other parameters apart from the base URL and credentials.


see usage section of pact provider verifier

https://github.com/pact-foundation/pact-provider-verifier#usage

The doc strings should be clear, that it is URL only, and we should not support the user providing a local file or a directory of pacts for a function called `verify_with_broker` - it is confusing, and I don't believe the scope of the users request.

If that is sorted, all I would say, add a note in readme, or show a verifier in the e2e example, performing a verification by URL in #332

pact/message_provider.py
sergeyklay: I'm not sure I understand what you mean. The current method signature (which I don't want to change in this pull request due this is offtopic) allows for an unlimited number of packets:

```py
def verify_pacts(self, *pacts, enable_pending=False, include_wip_pacts_since=None, **kwargs)
```

This has been the case for a long time and if it's wrong, it probably should be fixed, but it doesn't related to the current feature added. Isn't?

requirements_dev.txt
YOU54F: Thanks for the pointer and link, always yaks to be shaved 😅

sergeyklay:
Thank you for all your feedback on this pull request. I understand that there are many suggestions and requests for changes, but it seems that this PR is turning into a Frankenstein's monster and does not meet my quality standards. The initial objective of this pull request was to add a missing parameter to a single function, and I am not prepared to refactor half of the project in one PR.

In the near future, I will review all your comments, create a roadmap based on your suggestions, and gradually implement them. I am willing to create a new pull request addressing the current issue after thoroughly working through all your comments, including architectural changes, documentation format, and so on.

pact/message_provider.py
YOU54F: would a user not call this method `pact.Verifier.verify_pacts` if using a local pact file or dir

https://github.com/pact-foundation/pact-python/blob/4e3ca3858b8343a5057de506f19a4f3284bb8bd4/pact/message_provider.py#LL110C9-L110C15

and then call this method `pact.Verifier.verify_with_broker` if they were using a broker

https://github.com/pact-foundation/pact-python/blob/4e3ca3858b8343a5057de506f19a4f3284bb8bd4/pact/message_provider.py#LL118C9-L118C27

which wants to accept a singular pact_url

YOU54F: of these args, `enable_pending` & `include_wip_pacts_since` also don't make sense with regards to verifying via a broker url


        """Use Broker to verify.

        Args:
            broker_username ([String]): broker username
            broker_password ([String]): broker password
            broker_url ([String]): url of broker
            enable_pending ([Boolean])
            include_wip_pacts_since ([String])
            publish_version ([String])

        """
        
        
changed triggerd by by pact url

https://github.com/pactflow/example-provider/blob/master/src/product/product.consumerChange.pact.test.js

uses these options

```js
    const opts = {
      ...baseOpts,
      pactUrls: [process.env.PACT_URL],
      stateHandlers: stateHandlers,
      requestFilter: requestFilter
    };
```
        
        
one based on a provider change would involve a different set of parameters to a verifier

https://github.com/pactflow/example-provider/blob/master/src/product/product.providerChange.pact.test.js

```
    const fetchPactsDynamicallyOpts = {
      provider: 'pactflow-example-provider',
      consumerVersionSelectors: [
        { mainBranch: true },
        { deployed: true },
        { matchingBranch: true }
      ],
      pactBrokerUrl: process.env.PACT_BROKER_BASE_URL,
      // https://docs.pact.io/pact_broker/advanced_topics/pending_pacts
      enablePending: true,
      // https://docs.pact.io/pact_broker/advanced_topics/wip_pacts
      includeWipPactsSince: '2020-01-01'
    };

    const opts = {
      ...baseOpts,
      ...fetchPactsDynamicallyOpts,
      stateHandlers: stateHandlers,
      requestFilter: requestFilter
    };
```

There is some commonality between the modes the verifying with a broker

https://github.com/pactflow/example-provider/blob/master/src/product/pact.setup.js

YOU54F:
I think exposing a new function called verify_with_pact_url or something, that just takes the args required for that specific task might be more sensible, rather overloading the existing verify_with_broker function.

They are separate tasks

it should make it nice and easy to then have different sources

- by file (which we can do now)
- by consumer version selectors (which we can do now)
- by pact_url (which was being requested)

YOU54F:
That sounds sensible @sergeyklay sorry didn't mean to cause frustration

PR #341, opened 2023-04-14
Fix CORS parameter not doing anything

Lukas-dev-threads:
We found that pact ignores the CORS parameter of the `Consumer.has_pact_with()` function.
If the CORS parameter is set to True `--cors=*` is appended to the list of commands for the external mock service.

The mock service itself already supports this parameter.

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -214,6 +214,8 @@ class Pact(Broker):
             command.extend(['--sslcert', self.sslcert])
         if self.sslkey:
             command.extend(['--sslkey', self.sslkey])
+        if self.cors:
+            command.extend(['--cors'])
 
         self._process = Popen(command)
         self._wait_for_server_start()


pact/pact.py
```diff
@@ -214,6 +214,8 @@ def start_service(self):
             command.extend(['--sslcert', self.sslcert])
         if self.sslkey:
             command.extend(['--sslkey', self.sslkey])
+        if self.cors:
+            command.extend(['--cors=*'])
```
mefellows, line 218: Why the `=*`? I think just passing this without arguments (i.e. treat it as a flag) is all that is needed.

Or is this something specific to the way the python CLI lib interacts?

Lukas-dev-threads: I tested it without passing `=*` and it works fine.

The documentation of the mock server tells me that an argument must be provided

`pact\bin\pact\bin\pact-mock-service.bat help start`
```
 -o, [--cors=CORS]  # Support browser security in tests by responding to OPTIONS requests and adding CORS headers to mocked responses
```

But because `self.cors` is a boolean there is no way to pass different options for the cors parameter and it is
therefore better to append `--cors` without any value.

Corbie-42:
@mefellows we would also need this feature. Is this PR ready to be merged?

YOU54F (approved):

PR #346, opened 2023-05-03
chore(docs): update provider verifier options table

YOU54F:
- Table was mismatched, cli args were the python args and some missing
- updated links to examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,7 +486,7 @@ assert success == 0
 
 The parameters for this differ slightly in naming from their CLI equivalents:
 | CLI        |      native Python                      | notes |
-|-----------------|-------------------------------------------------------------------------------------------------|
+|-----------------|-------------------------------------------------------------------------------------------------|----|
 | `--log-dir`  | `log_dir`                                     ||
 | `--log-level`  | `log_level`                                     ||
 | `--provider-app-version`  | `provider_app_version`                                     ||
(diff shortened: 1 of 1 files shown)


sergeyklay (changes requested):
Thank you for your contribution to the project's documentation, it's greatly appreciated. However, before I can accept your pull request, I'd like to ask you to make a few changes:

1. Unfortunately, tables in markdown can often be a suboptimal way to present data. They can be difficult to read and edit, especially in an IDE. It's important that the table structure is as clear and easy to work with as possible. In particular, this means that all columns should be aligned. Please align the column widths based on the widest column to make it easier to read and edit the table for both users in a browser and developers in an IDE.
2. As for the relative links, since the content of this file will be rendered to HTML and displayed separately from its source code, relative links may or may not work properly. In any case, there are currently no relative links in the README.md file, and I would prefer not to introduce additional entropy. Please consider updating the links to absolute URLs where possible, to ensure consistent behavior.

Once again, thank you for your contribution, and I'll be glad to accept your pull request after these suggested changes have been made!

YOU54F:
>Unfortunately, tables in markdown can often be a suboptimal way to present data. They can be difficult to read and edit, especially in an IDE. It's important that the table structure is as clear and easy to work with as possible. In particular, this means that all columns should be aligned. Please align the column widths based on the widest column to make it easier to read and edit the table for both users in a browser and developers in an IDE.

agreed buddy, sorry didn't the edit via github's WYSIWIG editor, rather than in my IDE, shall rectify

> As for the relative links, since the content of this file will be rendered to HTML and displayed separately from its source code, relative links may or may not work properly. In any case, there are currently no relative links in the README.md file, and I would prefer not to introduce additional entropy. Please consider updating the links to absolute URLs where possible, to ensure consistent behavior.

Makes total sense to me buddy, consistency is king,  shall update to be absolute 👍🏾 will be tomorrow now as I am fading 😅

sergeyklay (approved):

PR #352, opened 2023-05-05
Revert/v2.0.0 standalone

YOU54F:
Hey!

It has been identified that there are a couple of issues noted in the new 2.x version of the pact-ruby-standalone, therefore I propose removing the changes applied in this PR

https://github.com/pact-foundation/pact-ruby-standalone/issues/101#issuecomment-1535926311

There was some additional changes to include testing across windows and linux on GitHub actions runners, we can move these to a separate pull request, to ensure that we test all all os's even with the 1.x version of the standalone.

For some reason I couldn't revert a merge commit, so this reverts each indiv commit.

Happy if someone can do a PR revert, to do that instead


YOU54F:
😂 ahhh all the failing tests.

YOU54F:
Going to switch to a draft and will investigate the failures when I get some more time, I noted we haven't released to pypi with these changes, so we are probably safe for a bit

PR #357, opened 2023-07-08
Feat/ffi - load full ffi namespace, initial ffi verifier

YOU54F:
Following on from the awesome work from @elliottmurray and @mikegeeves 

Will document properly over the next couple of days, (probably next week) 
- as part of the ffi maintainers guide
- pulling out other sources of pact verifications (both ruby, rust and jvm backed) so they can be referenced by others
- work out where any discrepancies lie
- where any logic is shared across client libraries and can be moved into the core
- quality of life improvements to any ffi methods to aid maintainers/library creators

Work done so far

- Loading of shared binaries for the following platform

| OS      | Architecture | 
| ------- | ------------ | 
| OSX     | x86_64       | 
| Linux   | x86_64       | 
| OSX     | arm64        | 
| Linux   | arm64        | 
| Windows | x86_64  |

- Full FFI namespace available by loading pact.h header file
- FFIVerify class created
- MessageProvider updated to support verification message pacts with rust core
- MessageProvider now returns same result signature as FFIVerify based on Verifier
- FastApi / FlaskAPI Provider's updated to verify with new FFIVerify
- Message Provider updated to verify with new FFIVerify
- gRPC v4 pact verification example based on area_calculator.proto used in https://github.com/pact-foundation/pact-plugins/tree/main/examples/gRPC/area_calculator
  - Pact was actually generated from this python grpc consumer example https://github.com/pact-foundation/pact-reference/blob/master/python/pact_plugin_grpc_v4.py in the pact-reference repo

## TODO

- Message metadata content-type currently hard coded to application/json
- Plugin tests when they fail, dont appear to return the failures or successes to the log buffer
- Log level gets globally applied, and isn't able to be changed, which interferes with the unit tests, and probably external tests within the same process. I think this is due to pact-reference using a global logging tracer.
- write migration guide
- bump to alpha version


YOU54F:
Not sure why Sync is complaining, I can't see the detail behind the check without logging into synk, and my gh username doesn't have access.

Should it be reporting back to the PR as a comment? 

🤔

YOU54F:
superseded by #367

PR #363, opened 2023-07-25
Feat/veri url

### LINKED ISSUE #332: No way to verify with a specific pact

YOU54F:
fixes #332 

provides the ability to

- use `verify` function to verify by pact Url(s) or local file(s)
- use `verify_with_broker` function to verify by consumer version selectors (dynamically fetched pacts)
- Allow support for MessageProvider to verify by pact Url(s) or local file(s), 
  - otherwise default to its existing behaviour of verifying by pactDir and constructing the path from the consumer/provider name
- update Flask Provider example and Message Provider example to demonstrate
  - verifying by file
  - verifying by url
  - verifying by consumer version selectors
- updated README to show verification options

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -21,10 +21,11 @@ pact/pact.py
 pact/provider.py
 pact/verifier.py
 pact/verify_wrapper.py
-pact/bin/pact-3.1.2.2-alpha-linux-arm64.tar.gz
-pact/bin/pact-3.1.2.2-alpha-linux-x86_64.tar.gz
-pact/bin/pact-3.1.2.2-alpha-osx-arm64.tar.gz
-pact/bin/pact-3.1.2.2-alpha-osx-x86_64.tar.gz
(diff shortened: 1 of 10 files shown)


YOU54F:
Hey @JP-Ellis 

It would still be relevant to support Pact Python being able to support verification by a pact url which might be a local file, or a remote url pointing to a pact on a broker.

This is a key element of our path to Pact Nirvana whereby the pact url would be used by provider verifications triggered by a consumer pact publish which requires verification.

Would love you to be able to either consider this or consider it in a more elegant way. Happy for this to be closed and a ticket to track :)

PR #313, opened 2022-10-17
feat: adds matchers to content

MdotMertens:
Hello,
Upon testing out pact I discovered that Message Pact does not support Matchers. 
See: https://github.com/pact-foundation/pact-python/issues/308 

This commit fixes this issue by checking whether the content contains any matchers and if it does, it converts the content

Issue #308, opened 2022-09-28
Matchers not working according to docs


MdotMertens:
Hello,
I want to use the matchers in my consumer tests but using them is kinda weird.
The way that I would expect them to work according to docs is:
```python
    event = {
        "event": "delivery:created",
        "delivery_type": matchers.Term("express|standard", "standard"),
    }

    (
        pact_no_publish.given("Delivery has standard shipping")
        .expects_to_receive("delivery:created")
        .with_content(event)
        .with_metadata({"Content-Type": "application/json"})
    )
```
What this does is to pass the Term object into the message.

What I have to do in this case is this:
```python
    event = {
        "event": "delivery:created",
        "delivery_type": matchers.Term("express|standard", "standard").generate()["data"]["generate"],
    }
```

This works but is the same as passing in a string directly and defeats the purpose of the matcher.

mefellows:
Thanks for the report, and that doesn't seem right to me at all. Are you able to provide a repro we can use to look at this?

P.S. `matchers.Term("express|standard", "standard")` - I think you are looking for an enum-like matcher here.

You probably actually want `matchers.Term("^express|standard$", "standard")` which will match the exact value. It's unlikely those words are going to be a subset of others, but you might want to know that for future matching.

MdotMertens:
Sure, here it is:
https://github.com/MdotMertens/python-pact-matchers

In it I reproduced the bug by using the Like matcher.
It passes the matcher into my consumer function instead of an actual value.

mefellows:
Ah, thanks! I see the problem.

In other languages, the event would have the matchers stripped. This can be done via the `reify` command on the `pact-message` command. 

![Screen Shot 2022-09-29 at 10 37 23 am](https://user-images.githubusercontent.com/53900/192912627-23041158-36ed-491e-a0f0-da4b311c8650.png)

```python
    with pact_no_publish:
        handler(event=event, context={})
```

I don't know Python very well, but ideally the `with pact_no_pubish` would provide the reified (body without matchers etc.) to the block, so that `handler` could be invoked with the cleaned JSON blob, and not with the matchers.

It should be fairly straightforward in the meantime to create a utility function that recurses the  `event` object and detects the presence of matchers, and extracts the value. Then your test should work as expected, and the matching rules will be properly serialised into the contract:

```python
    with pact_no_publish:
        reified_event = reify(event)
        handler(event=reified_event, context={})
```

[...]

PR #364, opened 2023-07-26
Mdot mertens feat/content matchers

YOU54F:
Updates to #313 

Adds examples relating to answer here in original raised issue https://github.com/pact-foundation/pact-python/issues/308#issuecomment-1455874464


JP-Ellis:
Do you think this would be covered as part of implementing the compatibility suites in:

- #420
- #421

YOU54F:
I reckon the following would be more apt for message pact 👍🏾 

- https://github.com/pact-foundation/pact-python/issues/380
- https://github.com/pact-foundation/pact-python/issues/379

The main point here is the user has to call `matchers.get_generated_values` to reify the message (strip it of matchers) before passing it to the handler. Other languages such as pact-js call the reify method under the hood so the user doesn't need to worry about the implementation detail. 

https://github.com/pact-foundation/pact-js/blob/768b70e4305669e71700a275ff4204f573788a7e/src/messageConsumerPact.ts#L193

PR #366, opened 2023-07-27
Feat/ffi consumer

YOU54F:
pact python consumer test examples using the PactFFI lib

- http pact
- message pact v3
- plug-in pact ( grpc ) v4

these are using direct calls to the ffi library so need wrapping in the Pact or Consumer dsl in pact-python

contributors welcome. this is your chance to propose a potential redesign to the pact python dsl if it isn’t quite right at the moment.

if not, hopefully connecting the dots should be a bit simpler :)

builds on #357 which introduces the ffi libs and verifier side now built out to use near enough the same dsl

tox doesn’t like my import of the area_calculator_client, it’s okay with pytest directly.


YOU54F:
superseded by #367

PR #367, opened 2023-08-02
Feature: PactV3 / VerifierV3 / MessageProvider interface (powered by pact_ffi)

YOU54F:
Following on from the awesome work from @elliottmurray and @mikegeeves 

Work done so far

- Loading of shared binaries for the following platform

| OS      | Architecture | 
| ------- | ------------ | 
| OSX     | x86_64       | 
| Linux   | x86_64       | 
| OSX     | arm64        | 
| Linux   | arm64        | 
| Windows | x86_64  |

- Full FFI namespace available by loading pact.h header file
- VerifierV3 class created
- PactV3 class created
- MatchersV3 class created
  - Term renamed to Regex 
- MessageProvider 
  - updated to support verification message pacts with rust core
  - returns same result signature as VerifierV3 based on Verifier 

- Examples 
  - Now executed in CI cross platform
  - support for a hosted pact broker
  - support for running without Docker (due to restrictions in CI systems)
  - support for using Pact CLI tools for publishing
  - Updated to use  PactV3 / VerifierV3 / MatchersV3
  - Added todo consumer/provider app with xml and binary payload example
    - [X] JSON
    - [ ] XML 
      - See issue https://github.com/pact-foundation/pact-reference/issues/305
    - [ ] Binary Payloads
  - gRPC v4 pact verification example added
    - based on area_calculator.proto used in https://github.com/pact-foundation/pact-plugins/tree/main/examples/gRPC/area_calculator
    - Pact was actually generated from this python grpc consumer example https://github.com/pact-foundation/pact-reference/blob/master/python/pact_plugin_grpc_v4.py in the pact-reference repo

Logging

- New options added for logging output, see [migration guide](https://github.com/YOU54F/pact-python/blob/feat/PactV3_VerifierV3/UPGRADE_FFI.md)

Docs

- Added a [migration guide](https://github.com/YOU54F/pact-python/blob/feat/PactV3_VerifierV3/UPGRADE_FFI.md)
- Added a TODO below and some scattered in code

## TODO

- [ ] Message metadata content-type currently hard coded to application/json
- [ ] Plugin tests when they fail, dont appear to return the failures or successes to the log buffer
- [x] Log level gets globally applied, and isn't able to be changed, which interferes with the unit tests, and probably external tests within the same process. I think this is due to pact-reference using a global logging tracer.
- update migration guide
- [x] bump to alpha version
  - `3.0.0b0`
- [ ] MessageConsumerV3
 - Created class, needs to call rust core rather than pact core
- [ ] MessageConsumerV4
- PactV4
  - Created basic example using `NativeMockServer`
  - Needs Consumer DSL
- [ ] Create a proper native_mock_server interface that doesn't have any logic in it
  - I've just imported required methods at the moment. 
  - Should native_mock_server should be a direct bridge to the FFI
  - PactV3 currently uses NativeMockServer but it quite aware of the logic requirements of the FFI
- [ ] Error handling on the native interop side
  - Generally decent through python's `cffi` module with somewhat descriptive messages and print out of the objects it was trying to marshall.
- [ ] pactffi_string_deletes
  - Free anything that needs to be freed 
- [ ] matcher/generators
  - [x] Matchers in body 
  - [ ] Matchers in path
  - [ ] Matchers in queries
  - [ ] Matchers in headers
- [ ] consumer side lifecycle
  - run mock_server_matched in users tests after they have issued the request and before they start doing their unit test assertions
   - allow them to then call a finalise to write the pact file to disk only after these assertions are complete
   - Currently this is all in one with the existing signature `provider.verify` which also tears down the mock server for that interaction
   
## V2 -> V3 notes

1. http_proxy returns `{contents: message}` but required to be `message` in rust core

## How can I test it?

1. Clone the repo and follow the setup instructions.

or add the following to pull from my latest beta branch

`'pact-python @ git+https://github.com/you54f/pact-python@3.0.0b0'`


YOU54F:
Bit of recap.

I've started testing this out by importing into other projects in the wild.

1. I've got issues with top level imports, `pact.ffi.x` isn't exposed in the `pact` import and I'm not sure if I want to expose the ffi modules to the end user.

It works on my machine, and when pip installed from another repo, but not when downloaded from a release tarball.

2. Content-type header must be explicitly set on req/res body. I am using format `[{"name":"content-type","value":"application/json"}]`
  - It's verbose
  - It doesn't match a python Headers object, which some of the examples in the wild I found.

YOU54F:
Closing and new work will be tracked under #396 🎉

PR #368, opened 2023-09-07
chore: update examples to work with podman on macos

YOU54F:
Testing out the examples on MacOS with Podman, over Docker compose

```console
brew install podman
podman machine init
podman machine start
export DOCKER_HOST='unix:///Users/yousaf.nabi/.local/share/containers/podman/machine/qemu/podman.sock'
make examples
```


YOU54F:
No longer required, was using to help get @JP-Ellis setup locally

Issue #369, opened 2023-09-08
Migrate to `pyproject.toml`


JP-Ellis:
Following [PEP 518](https://peps.python.org/pep-0518/) and [PEP 621](https://peps.python.org/pep-0621/), we should migrate the older `setup.cfg` and `setup.py` build system to one making use of `pyproject.toml`. Furthermore, the use of `setup.py` is strongly discouraged and many aspects will be discontinued:

> Setuptools offers first class support for `setup.py` files as a configuration mechanism.
>
> It is important to remember, however, that running this file as a script (e.g. `python setup.py sdist`) is strongly **discouraged**, and that the majority of the command line interfaces are (or will be) **deprecated** (e.g. `python setup.py install`, `python setup.py bdist_wininst`, …).
>
> We also recommend users to expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the `setup.py` minimal with only the dynamic parts (or even omit it completely if applicable).
>
> See [Why you shouldn’t invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) for more background.

<div align="right"><a href="https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setuppy-discouraged">source</a></div>

The existing `setup.py` build script also makes use of [`distutils`](https://docs.python.org/3/library/distutils.html) which has been deprecated since 3.10 and is scheduled to removal in 3.12.

As part of this change, we also have an opportunity to change the underlying build system that handles the packaging of the Python package into a distributable bundle. The main ones to consider are:

- [Setuptools](https://setuptools.pypa.io/en/latest/)
- [Poetry](https://python-poetry.org/)
- [Maturin](https://github.com/PyO3/maturin)
- [Hatch](https://hatch.pypa.io/latest/)
- [PDM](https://pdm.fming.dev/latest/) and [PDM-backend](https://pdm-backend.fming.dev/)
- [Flit](https://flit.pypa.io/en/stable/index.html)

An excellent write up by Jay Qi is available in their blog post [The Basic of Python Packaging in Early 2023](https://drivendata.co/blog/python-packaging-2023).

As `pact-python` will eventually make use of the Pact Rust core, any build system considered must be able to support so-called 'extensions' (which are typically used to build C extensions).

# Deep Dive

[...]

Issue #350, opened 2023-05-03
Drop support for Python < 3.7


sergeyklay:
I would like to propose dropping support for Python versions below 3.7 in our project. Active development of Python 3.6 ended over 6 years ago (22 December 2016), and since then, some of libraries that we use began to drop support of older Python versions, including Python 3.6.

Main problems of supporting Python < 3.7:

- **Outdated code:** Supporting older Python versions may lead to the use of outdated and less optimal solutions.
- **Increased development and debugging time:** Supporting older Python versions requires extra effort as we need to take into account the peculiarities of each supported version.
- **Library compatibility issues:** Many modern libraries no longer support older Python versions, which may create difficulties in their usage and integration.
- **Missing new features:** Older Python versions do not support some of the new features and improvements that have been introduced in more recent versions, such as f-strings, type annotations, data classes, and others.
- **Decreased performance:** Supporting older Python versions may reduce the project's performance as some optimizations are only available in newer versions.
- **Complicated testing:** Maintaining (including tricks with CI/CD, Docker support, and local testing) across exotic Python versions requires more time and resources.

Considering the aforementioned problems, it is recommended to drop support for Python versions below 3.7, which will allow us to improve code quality, speed up development, and reduce the risk of errors arising from the use of outdated features and libraries.

/cc @mikegeeves

YOU54F:
rationale makes total sense, we will be dropping node 14 in pact js and pact js core as they went eol last week.

we’ve just been upgrading the ruby codebases for 3.2 

checking python eol, python 3.7 drops support for in 1 month and a week 

https://endoflife.date/python

do we consider tabling that aswell, as part of this or a further motion.

i am all for supporting and advocating our users upgrading to the latest versions, i appreciate there are cases where they may not be able to, but as  our tools are open source they can maintain older versions themselves if required

sergeyklay:
Yes, 3.7 will lose active support very soon. All I want from the resolution of this issue is to stop supporting at least 3.6. And we can get rid of 3.7 later.

PR #371, opened 2023-09-12
chore: migrate to pyproject.toml and hatch

JP-Ellis:
This PR overhauls the underlying build process for the library.

## Background

The previous build process relied on `distutils` which is due to be deprecated in 3.12. As a result, it is necessary for us to look at updating the build system.

Furthermore, the use of `setup.py` is now discouraged, and the preference is to declare as much of a package's information in `pyproject.toml`.

## Choice

The choice to migrate to `hatch` was made for the following reasons:

- It offers a very simple management of the venv. No more need to `python -m venv .venv` and `pip install`, `hatch` handles all of that automatically when creating the virtual environment.
- `hatch` supercedes `tox`, allowing for multiple python versions to be tested in a single command.
- `hatch` manages the build process, and offers a nicer way to hook in a custom build process to download the `pact` standalone binaries.

A minor change to the packaging of the library now places the binaries in `pact/bin` instead of `pact/bin/pact/bin`. The `constants.py` file has been accordingly updated to reflect this change in case anyone was making direct use of the binaries.

While this change is rather significant, it should not affect the end user experience. Users will still be able to `pip install pact-python` from PyPI. Other than for the aforementioned, there has been no changes to the library code.

## CI/CD Changes

This PR also includes a commit migrating the CI/CD pipelines to make use of Hatch, and now builds the wheels using the [pypa/cibuildwheel](https://github.com/pypa/cibuildwheel) GitHub action. This action handles most of the logic to build and distribute binary wheels for all supported Python versions and platforms. The current list of wheels built are:

- `cp{38,39,310,311,312}` on:
  - `macosx_10_16_arm64.whl`
  - `macosx_10_16_x86_64.whl`
  - `manylinux_2_17_arch64`
  - `manylinux_2_17_x86_64`
  - `musllinux_1_1_x86_64`
  - `win32`
  - `win_amd64`
- `pp{38,39,310}`
  - `win_amd64`
  - `macosx_10_16_x86_64`
  - `manylinux_2_16_x86_64`

_pypy only has support up to Python 3.10, and only recently has supported for arm64 build_

Resolves: #350
Resolves: #369
Resolves: #389
Refs: https://docs.python.org/3/whatsnew/3.10.html#distutils-deprecated
Refs: https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setuppy-discouraged


pyproject.toml
```diff
@@ -0,0 +1,156 @@
...
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3 :: Only",
+    "Topic :: Software Development :: Testing",
+]
+
+requires-python = ">=3.8,<4.0"
+dependencies = [
+    "click      ~= 8.1",
+    "fastapi    ~= 0.103",
```
mefellows, line 32: I noticed this and it seems equivalent to the existing requirements, however is this a dev-only dependency or does Hatch not distinguish between these during build, and figures out what is required automatically in the distribution step?

mefellows (approved):
It looks good to me, but I'm not qualified to properly understand some of these changes. A review/pass through from somebody who knows Python well would be ideal.

pyproject.toml
JP-Ellis: These are actually dependencies of the library itself (not just dev dependencies). See for example:

https://github.com/pact-foundation/pact-python/blob/ed5f86c95ae0856338f0fd244c06399ce395f353/pact/cli/verify.py#L6

Upon review, some of them do appear to be dev dependencies (e.g., `fastapi` is only used within the examples). Let me double check

mefellows: Ah, it's used here: https://github.com/pact-foundation/pact-python/blob/ed5f86c95ae0856338f0fd244c06399ce395f353/pact/http_proxy.py#L4

I thought it was a different web server implementation, but this clarifies. Thanks!

.github/workflows/build.yml
```diff
@@ -0,0 +1,171 @@
...
+        path: ./wheelhouse/*.whl
+        if-no-files-found: error
+
+  build-arm64:
+    name: Build wheels on ${{ matrix.os }} (arm64)
+    runs-on: ${{ matrix.os }}
+    # As this requires emulation, it's not worth running on PRs
+    if: >-
+      github.event_name == 'push' &&
+      startsWith(github.event.ref, 'refs/tags')
```
JP-Ellis, line 93: This has been tested on GitHub Actions and is confirmed to work, along with the following `check` step.

JP-Ellis:
@YOU54F, @sergeyklay and @mikegeeves I understand that this is a pretty substantial change to a foundational part of this project. Happy to organise a chat if it is easier to talk through this review.

YOU54F:
> @YOU54F, @sergeyklay and @mikegeeves I understand that this is a pretty substantial change to a foundational part of this project. Happy to organise a chat if it is easier to talk through this review.

Yeah I'm up for that, I'm pretty flexible on times

YOU54F:
do we currently publish one package which contains all the pact-ruby-standalone archives for each platform, and this will allow us to create platform/arch specific packages, therefore distributing smaller packages to end users?

If so that will be nice!

I note `musllinux_1_1_x86_64`

We have some guidance on Alpine, and Pact, which differs depending on the Pact backed (Ruby vs Rust)

some of the docs are [here](https://docs.pact.io/docker#alpine-linux)

TL;DR

- pact-ruby core distributed with pact-ruby-standalone, via traveling-ruby provides a glibc flavour of Ruby, therefore Alpine users need gcompat libs
- pact-rust core
  - we provide musl .a binaries 
   - https://github.com/pact-foundation/pact-reference/releases/tag/libpact_ffi-v0.4.8
   - however I don't think many, (if any) other projects are using these, some comments here from my adventures building them and Ron's comments [on usage by end users](https://github.com/pact-foundation/pact-reference/pull/276#issuecomment-1552209563)

.github/workflows/build.yml
YOU54F: How slow is slow? ahh just seen 40 ish mins

Screenshot 2023-09-14 at 13 18 16

JP-Ellis: In terms of the computation, you are correct, nothing is _too_ computationally complex.

I make use of [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) to manage the building of wheels against the various targets. It does handle a couple of extra things on top of just building the wheel (it can run the test suite, or in the above case, just tests a rather trivial use of Pact), it audits the wheel, and for binary wheels, also ensures that all necessary system libraries are packaged as well. It also repeats this for all main versions of Python.

I suspect that the GitHub Runners are quite highly optimised for running just Linux, and when you have a containerised environment with `qemu` handling the architecture virtualisation, things evidently slow down quite drastically.

The 50 minute runtime ultimately is not _awful_, but I certainly wouldn't want to force every PR to wait on this to succeed.

pyproject.toml
```diff
@@ -0,0 +1,157 @@
...
+classifiers = [
+  "Development Status :: 5 - Production/Stable",
+  "Environment :: Console",
+  "Intended Audience :: Developers",
+  "License :: OSI Approved :: MIT License",
+  "Operating System :: MacOS :: MacOS X",
+  "Operating System :: Microsoft :: Windows",
+  "Operating System :: POSIX :: Linux",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.8",
```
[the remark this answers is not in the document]
JP-Ellis: I put Python 3.6 and 3.7 support in the back of my mind temporarily, as I wanted to get Hatch to work first. I am happy to add them back in if that is really controversial, but my opinion is we should not support them.

To be more specific:

- Nothing in the library code has changed, and therefore I don't expect 3.6 and 3.7 to cause any issues with the upgrade. If someone really needed to support these versions, we could add this back in as required.
- 3.6 and 3.7 are no longer maintained by Python and therefore any security vulnerabilities discovered now will remain.
- The main issue is that a lot of dependencies required by pact-python themselves do not support 3.6 and/or 3.7. This resulted in effectively two separate dependency definitions: one set for supported versions of Python, and another set with versions pinned to historical versions. In addition, a lot of these older pinned versions have security vulnerabilities which we would require end-user t take on.

If there's no objections, I'll mark this PR as resolving #350 as well :)

JP-Ellis:
> do we currently publish one package which contains all the pact-ruby-standalone archives for each platform, and this will allow us to create platform/arch specific packages, therefore distributing smaller packages to end users?
> 
> If so that will be nice!

We are indeed currently publishing one (rather large) package which contains all `pact-ruby-standalone` archives.

The new build process with `hatch` target a specific platform, separates this out and makes sure to package only the binaries relevant to the target platform.

> I note `musllinux_1_1_x86_64`
> 
> We have some guidance on Alpine, and Pact, which differs depending on the Pact backed (Ruby vs Rust)
> 
> some of the docs are [here](https://docs.pact.io/docker#alpine-linux)
> 
> TL;DR
> 
> * pact-ruby core distributed with pact-ruby-standalone, via traveling-ruby provides a glibc flavour of Ruby, therefore Alpine users need gcompat libs
> * pact-rust core
>   
>   * we provide musl .a binaries
>   * https://github.com/pact-foundation/pact-reference/releases/tag/libpact_ffi-v0.4.8
>   * however I don't think many, (if any) other projects are using these, some comments here from my adventures building them and Ron's comments [on usage by end users](https://github.com/pact-foundation/pact-reference/pull/276#issuecomment-1552209563)

When it comes to the Rust core, we should be able to create wheels that target musl linux quite straightforwardly.

As for the Ruby core at the moment, as there are no upstream musl binaries generated, the build process didn't work. I can look later whether we want to add a compatibility layer, but I think that should be reserved to a separate PR.

.cirrus.yml
```diff
@@ -1,51 +1,49 @@
...
     - brew install pyenv
     - pyenv install ${PYTHON}
     - pyenv global ${PYTHON}
     - pyenv rehash
-  ## To install rosetta
-    # - softwareupdate --install-rosetta --agree-to-license
-  << : *BUILD_TEST_TASK_TEMPLATE
+    - python -m pip install --upgrade pip pipx
+    - pyenv rehash
+    - pipx install hatch
```
YOU54F, line 47: do we always want to run against the latest version of hatch, or do we want to pin this so it could be picked up via say a dependabot config (or similar) to ensure we can test version updates and ensure consistency.

YOU54F: as a note, my first experience was that I needed to install hatch, which is where I came back to the CI scripts to see how we are installing it :)

```
pact-python on  feat/pyproject@origin:feat/pyproject via 🐳 desktop-linux via 🐍 v3.11.4 on ☁️  (eu-west-2) 
🕙15:02:03 ❯ make                                                

  clean      to clear build and distribution directories
  examples   to run the example end to end tests (consumer, fastapi, flask, messaging)
  consumer   to run the example consumer tests
  fastapi    to run the example FastApi provider tests
  flask      to run the example Flask provider tests
  messaging  to run the example messaging e2e tests
  package    to create a distribution package in /dist/
  release    to perform a release build, including deps, test, and package targets
  test       to run all tests


pact-python on  feat/pyproject@origin:feat/pyproject via 🐳 desktop-linux via 🐍 v3.11.4 on ☁️  (eu-west-2) 
🕙15:02:40 ❯ make clean
hatch clean
make: hatch: No such file or directory
make: *** [clean] Error 1

pact-python on  feat/pyproject@origin:feat/pyproject via 🐳 desktop-linux via 🐍 v3.11.4 on ☁️  (eu-west-2) 
🕙15:02:44 [🔴 USAGE] ❯ make package
hatch build
make: hatch: No such file or directory
make: *** [package] Error 1

pact-python on  feat/pyproject@origin:feat/pyproject via 🐳 desktop-linux via 🐍 v3.11.4 on ☁️  (eu-west-2) 
🕙15:02:51 [🔴 USAGE] ❯ make test   
hatch run all
make: hatch: No such file or directory
make: *** [test] Error 1
```

hatch_build.py
```diff
@@ -0,0 +1,147 @@
...
+                machine = "arm64"
+            else:
+                msg = f"Unknown Linux machine {platform}"
+                raise ValueError(msg)
+
+            url = PACT_URL.format(version=version, os=os, machine=machine, ext="tar.gz")
+
+        else:
+            msg = f"Unknown platform {platform}"
+            raise ValueError(msg)
```
YOU54F, line 100: nice

hatch_build.py
YOU54F, line 1: Looks sane and easy to understand, and not too dissimilar from the previous setup

It is far quicker however, very impressed with the speed

pyproject.toml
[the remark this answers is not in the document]
YOU54F: Right I'm down with that. If users need to support older versions of Python for their application, they are welcome to propose backports. Should we consider snap shotting the branch prior to merging this.

I suppose only speaking to you yesterday, we could support 3.6/3.7 via hatch, ergo meaning a user could if required, so the snapshot possibly isn't required.

Makefile
```diff
@@ -105,34 +96,11 @@ examples: consumer flask fastapi messaging
...
-	@echo "\n$(green)Make a .venv dir$(sgr0)"
-	~/.pyenv/versions/${PYENV_VERSION}/bin/python3 -m venv ${CURDIR}/.venv
-
-	@echo "\n$(green)Make it 'available' to pyenv$(sgr0)"
-	ln -sf ${CURDIR}/.venv ~/.pyenv/versions/${PROJECT}
-
-	@echo "\n$(green)Use it! (populate .python-version)$(sgr0)"
-	pyenv local ${PROJECT}
+test:
+	hatch run all
```
YOU54F, line 104: fails on lint, as noted in the ci scripts

YOU54F:
```console
🕙15:18:21 ❯ hatch run test:all
───────────────────────────────────────────────────────────────────── test.py3.8 ─────────────────────────────────────────────────────────────────────
/bin/sh: tests: command not found

```

PS not sure why its saying `py3.8`, my currently installed versions are as below. Is that failing because I don't have python 3.8.

```
❯ pyenv versions 
  system
  2.7.18
* 3.11.4 (set by /Users/yousaf.nabi/.pyenv/version)

```

`hatch run test` works grand and all the unit tests passing. can see my wheel built the dist folder.

How would one directly install the egg in an external project to test?

.github/workflows/build.yml
YOU54F: 40/50 mins is similar to what I see using cross (qemu) and docker building ruby from source, cross arch.

I'm surprised to see it take that long, maybe that is because the audit and stuff also running under emulation.

Anyhoo, as we won't build these that often its probably of no consequence

```console
🕙15:32:53 [🧱 INT] ❯du -sh dist/*
 18M    dist/pact_python-2.0.1-cp311-cp311-macosx_10_16_arm64.whl
 60K    dist/pact_python-2.0.1.tar.gz
 ```
 
File sizes are much nicer now with the indiv packaged wheels. But I assume the pure python stuff in the smaller package, and the pact-ruby-standalone is embedded in the whl. I still can't fathom why it takes so long to do python arm64 things.

anyway, its other peoples machines (ie github runners) so its fine!

YOU54F:
RE: Musl packages

Here is the current advice for alpine users with the ruby packages

https://github.com/pact-foundation/pact-ruby-standalone/wiki/Using-the-pact-ruby-standalone-with-Alpine-Linux-Docker

I not sure if that helps in getting the neccessary deps in place to build a `musl` wheel, even if its a) just packages the glibc backed linux binaries and we need to tell users to install the neccessary pre reqs with apk or b) if hatch is smart enough to do that for users.

Either way the acid test is just being able to run a pact test in an alpine docker, as you can today, albeit with our workarounds

we already have an example docker alpine file here

https://github.com/pact-foundation/pact-python/blob/master/docker/Dockerfile

I don't think its a show stopper for me though, I would much rather tackle the alpine problem with you for pact-reference/pact_ffi and its consumption in users projects to try and make life on Alpine a possibility with the rust core.

and if neccessary we can circle back on ruby, 

I have been tickling some alternative ruby packagers which provide single file executables, and musl based builds

https://gist.github.com/YOU54F/3775e66e6090e0371c11601e6b75c305

and got enough knowledge of the traveling ruby system to build musl based rubies

YOU54F (approved):
I'm okay with this, I think we might want to do a major bump if dropping python versions though.

Unless the convention is minor in Python land. ( I had a bit of questioning around this in Ruby land the other day) See [thread](https://github.com/pact-foundation/pact-provider-verifier/issues/108#issuecomment-1715689419)

also consideration given in the docs for whatever the outcome is for alpine users

Makefile
JP-Ellis: Within hatch, I have defined two sets of venvs. The default venv is the one used interactively by hatch during regular development work, and will use whatever (compatible) Python you have as default.

I have also defined a `test` environment which is itself a matrix of Python version:

https://github.com/pact-foundation/pact-python/blob/324a624149e909ee8174537407db3d9d190d640e/pyproject.toml#L115-L121

When running `hatch run test:all`, it should setup a venv for each Python version in the matrix and run the test suite. Hatch, to my knowledge, automatically discovers Python versions in your system.

If you don't have Python 3.8 at all, I don't know how this ran for you. Do you have it installed separately to `pyenv`, perhaps with `brew install python@3.8` or maybe `conda` if it supports this? I don't believe hatch will automatically download and install versions of Python automatically, but I might be wrong.

YOU54F: ahh yeah there will more than likely be a python@3.8 required from various homebrew packages, not managed by pyenv

JP-Ellis:
Regarding the dropping of Python version, I'll reword the commit that introduces the change and make it clear it is a `BREAKING CHANGE` (ensuring it follows conventional commits). Unfortunately, this might automatically cause your reviews to go stale so just a heads up :)

In my experience, and based on the way dependencies have done this too, a minor version upgrade is sufficient in Python when changing the supported versions of Python.

.cirrus.yml
JP-Ellis: Good question. I think installing the latest version of hatch should generally not be an issue. Hatch is not an actually dependency of `pact-python`, and instead is just a CLI interface and a build backend. I would be very surprised if regular updates to hatch broke this.

Of course, should hatch come out with a completely overhaul in `v2`, then we might need to pin it then, but I'm happy to leave this as future problem.

In the same vein, I'm of two minds as to whether we should update system dependencies (`apt` and `brew`) as part of the workflow. Personally, I typically lean towards faster CI/CD builds and therefore use whatever packages are installed in the image, and then trust the CI/CD provider to regularly update these packages.

Lastly, would you like me to add a `make` command to install hatch? Or not really worth it?

YOU54F: >Of course, should hatch come out with a completely overhaul in v2, then we might need to pin it then, but I'm happy to leave this as future problem.

Yeah I'm okay with that, the CI will catch any changes, and alert, and we can action accordingly

>In the same vein, I'm of two minds as to whether we should update system dependencies (apt and brew) as part of the workflow. Personally, I typically lean towards faster CI/CD builds and therefore use whatever packages are installed in the image, and then trust the CI/CD provider to regularly update these packages.

You get a more repeatable build by not pulling in new updates each version, I've always just updated as a matter of course, going from practises used in Docker files, but also find it incredibly annoying when building things locally that link dynamically, and the dep has been updated with a new named shared lib, causing my built programs to fail, until they are rebuilt. Also why I like statically compiled executables, that will always work, even if they are chonkier.

> Lastly, would you like me to add a make command to install hatch? Or not really worth it?

could do, if we did I would favour it being installed with python, so it saves x platform package manager concerns

if we don't, just have it as a pre-req for development somewhere in the docs.

you could have a make doctor command, which checks if the hatch executable is available and alerts people on how to install it

YOU54F:
> In my experience, and based on the way dependencies have done this too, a minor version upgrade is sufficient in Python when changing the supported versions of Python.

Awesome thanks for confirming

>Unfortunately, this might automatically cause your reviews to go stale so just a heads up :)

no worries 👍🏾 thanks for the heads up

sergeyklay:
The CONTRIBUTING.md file seems to have remained unchanged and still contains version 3.6.

YOU54F:
👋 Hey @sergeyklay 

PR #390 contains updates to the contributing document, removing the mention of the python versions altogether ( i've just reread the updated doc proposed )

it's probably worth a mention in that PR, as to where a user can see the versions in use ( the toml file and ci are possibly better sources of truth and could have that information on the main readme - supported python versions - as it's relevant not just for developers but end users) in which case it might be additional duplication having it in contributing guidelines.

we can just say check the readme supported versions section or check the ci matrix/pyproject.toml

JP-Ellis:
In most cases, I think people don't really care which version of Python is supported other than "the current one". If someone is looking for it, I think there are really two main places to look for this:

- The first and I think most frequent is on PyPI. For example, the current page for Pact Python has:

  image

- The 2nd place I think is typically as badges on the README.md (which typically also propagates to PyPI). A good example is the Hatch README:

  image

As for this repository, we have the former covered (it is in the new `pyproject.toml`), and I would like to updated to README eventually, but I think this will only be published with version 3 of this library.

PR #387, opened 2023-09-15
Fix existing examples

### LINKED ISSUE #370: Fix existing examples
The existing [examples](https://github.com/pact-foundation/pact-python/blob/33352219fdc9fef3940cfffac3bdf06c833bdd84/examples) seem a bit fragile and required me to make some changes to the underlying code to get them to work. The barrier to getting them to work should be decreased to help facilitate adoption.

Some of the changes to look at:

- [x] Fix issues with the shared fixture not being able to publish the (seemingly inexistent) Pact file
- [x] Reuse the pact broker between the examples, instead of spinning it up and down every time
- [x] Review whether we need to support SSL for this example
- [x] See whether it should be integrated with the broader `pytest` suite (perhaps gated behind a flag?)

      Opted to just have them in a separate directory. `pytest examples/` (or `hatch run examples`) will suffice

JP-Ellis:
This PR resolves #370.

When first picking up `pact-python`, I immediately ran into some issues with running the examples.

This PR migrates the examples into a single test suite to be executed with `hatch run example`. This will (optionally) spin up the pact broker container, and maintain it across all consumer and provider examples.

The idea for this example is to have the consumer run first and publish the contracts, and then to verify the providers against these contracts.

In doing so, I have also attempted to document rather explicitly the way the examples are written and how they might be used. It is not designed to replace the docs in any way, but hopefully will provide a much more hand-lead introduction for `pact-python`.


pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+[project.optional-dependencies]
+types = ["mypy ~= 1.1", "types-requests ~= 2.31"]
+test = [
+  "coverage[toml] ~= 7.3",
+  "httpx          ~= 0.24",
+  "mock           ~= 5.1",
+  "pytest         ~= 7.4",
+  "pytest-cov     ~= 4.1",
+  "testcontainers ~= 3.7",
+  "yarl           ~= 1.9",
```
YOU54F, line 58: How do I install these optional-deps

I've already ran 

```console
hatch run example 
ImportError while loading conftest '/Users/yousaf.nabi/saf_stuff/dev/pact-foundation/pact-python/examples/conftest.py'.
examples/conftest.py:15: in <module>
    from yarl import URL
E   ModuleNotFoundError: No module named 'yarl'
```

managed to get pip to install with `pip install -e '.[test]'` 

but still get the same error as above

JP-Ellis: Hatch should installed these dependencies automatically when it first creates the virtual env. When dependencies change, hatch _should_ automatically detect this and (un)install the changed dependencies; but I have encountered this issue myself once. When I re-created the venv though, all dependencies were correctly installed.

I suspect this is an edge case whereby hatch does not detect transitive dependencies. In particular, there is a chain `pact-python[dev] -> pact-python[test] -> yarl` and `pact-python[dev]`, and the default venv should install all `[dev]` dependencies:

https://github.com/pact-foundation/pact-python/blob/7279cac38ce15e15bb95ef364ea14f829d2e6e07/pyproject.toml#L101-L103

I'll check whether this is a known issue upstream.

YOU54F: Thank you! I think I was a bit early to the party for a review, but good to see how its going to play out!

YOU54F:
docker not available on macos runners in cirrus ci or github actions.

licensing reasons on gha and because of apple virtulization restrictions for cirrus ci, which doesn't permit nested virtualization.

I took a trip down ruby executable package lane, to look at packaging up the pact broker, as part of our pact-ruby-standalone bundle.

It worked well for linux and macos, but for windows, is a no go, as it relies on natives gems for the pact broker, of which the traveling-ruby package doesn't provide.

There are some windows projects that support packaging windows apps, with native extensions

https://github.com/pact-foundation/pact-ruby-standalone/issues/118

and also package into a single file.

I thought it would be nice for macos and linux as well.

That lead me to fixing up ruby-packer

https://github.com/pmq20/ruby-packer/pull/177

Mmm that was okay but there were issues loading native gems, which cropped up occasionally and I couldn't pin it down, so ended up thinking, why not try to roll my own. 

In doing so, I stumbled across a random SO post here

https://stackoverflow.com/questions/76579841/package-standalone-rubygems-for-a-macos-app/76741361#76741361

I documented all my findings here

https://gist.github.com/YOU54F/3775e66e6090e0371c11601e6b75c305

note if we got some hardware for self hosted runners, we could runner docker natively on mac, and not be using macos in a vm (with cirrus)

YOU54F:
I also made changes in my PR to support switching between using a hosted OSS broker that PactFlow provides (for ci/testing/poking around)

https://github.com/pact-foundation/pact-python/pull/367/commits/59e2ef6c3995cddc472ae0db6151ff5a3d144d95

I was hitting rate limits as the CI tests would run against x python versions on 3 OS's, and 2 arches,  whilst using fix app version and branches, and the same consumer/provider names.

We would be limited cos we are smashing the same endpoint with the same data xD

adding the commit and branch from the actual branch/commit of the repo along with an identifier for <os>-<arch>-<ci_system> would be useful to avoid clashing, rate limiting and identification, back, in case of discrepancies between any of the generated artefacts (pact files in the examples case)

pyproject.toml
YOU54F: ahh okay, so its happier now i've done the following  (been able to find yarl and at least run the examples)

```
git clean -fdx
hatch env remove
hatch env create
```

YOU54F:
Odd error in CI on Cirrus for MacOS and Linux

```console
_multicall
    res = hook_impl.function(*args)
  File "/tmp/cirrus-ci-build/examples/conftest.py", line 24, in pytest_addoption
    action=argparse.BooleanOptionalAction,
AttributeError: module 'argparse' has no attribute 'BooleanOptionalAction'
```

seems to relate to

https://nono.ma/attribute-error-module-argparse-has-no-attribute-boolean-optional-action

odd how we don't see it in github actions! oh right this will make sense, we aren't running the examples on anythign other than the latest version, whereas we are running it across all the versions in cirrus

https://github.com/pact-foundation/pact-python/actions/runs/6233729823/workflow#L56

YOU54F:
If I say as an end user, lift and shift an example, into my own repo (which for arguments sake is empty), how would I easily be able to work out what my deps look like for any particular example?

Screenshot 2023-09-19 at 19 56 06

Could we solve that in a separate way, if the problem statement is, as a user, I easily want to replicate one of the python e2e examples as a bootstrap in a new repo or just on my local machine, could pact-python help me bootstrap that.

(or thinking wider, could our pact languages more generally do this, so its actually part of the core tool, maybe its a way of extensions, people can bring their own templates in, but we can use it for having our examples close to our working code in each respective language,  having a core cli tool that can interrogate a repo and work out what example templates are available)

not something for this pr just food for thought

JP-Ellis:
Wow thank you for all the comments!

> docker not available on macos runners in cirrus ci or github actions.
> 
> licensing reasons on gha and because of apple virtulization restrictions for cirrus ci, which doesn't permit nested virtualization.

That's fine. I can disable these checks on Cirrus. As for GitHub Actions, I'm pretty sure I've used Docker images before or some other sort of containerisation/virtualisation (it's needed to build the wheels targetting `arm64`) 🤔

Personally, I'm fine with testing the examples against only a subset of the targets. The test suite should definitely be validated against _all_ supported versions, but the examples are more pedagogical. The `BooleanOptionalAction` error is in fact because it was introduced in Python 3.9.

> I also made changes in my PR to support switching between using a hosted OSS broker that PactFlow provides (for ci/testing/poking around)
> 
> [59e2ef6](https://github.com/pact-foundation/pact-python/commit/59e2ef6c3995cddc472ae0db6151ff5a3d144d95)
> 
> I was hitting rate limits as the CI tests would run against x python versions on 3 OS's, and 2 arches, whilst using fix app version and branches, and the same consumer/provider names.
> 
> We would be limited cos we are smashing the same endpoint with the same data xD
> 
> adding the commit and branch from the actual branch/commit of the repo along with an identifier for --<ci_system> would be useful to avoid clashing, rate limiting and identification, back, in case of discrepancies between any of the generated artefacts (pact files in the examples case)

Thanks! I did add the option in the examples allowing for `--broker-url` to be specified for this purpose. It's also useful if I'm running the broker locally and want to keep it running after the test, as opposed to having the test suite spinning up the container and shutting it down every time.

> If I say as an end user, lift and shift an example, into my own repo (which for arguments sake is empty), how would I easily be able to work out what my deps look like for any particular example?
> 
> Screenshot 2023-09-19 at 19 56 06
> Could we solve that in a separate way, if the problem statement is, as a user, I easily want to replicate one of the python e2e examples as a bootstrap in a new repo or just on my local machine, could pact-python help me bootstrap that.
> 
> (or thinking wider, could our pact languages more generally do this, so its actually part of the core tool, maybe its a way of extensions, people can bring their own templates in, but we can use it for having our examples close to our working code in each respective language, having a core cli tool that can interrogate a repo and work out what example templates are available)
> 
> not something for this pr just food for thought

That's a good question, and the last commit (still a WIP) for this PR is going to check all of the documentation and make sure it is clear. This way, they can 'lift-and-shift' what is relevant to them.

I think from a pedagogical perspective, it is important to show the whole process at once, especially in a very simple example. If I, as a developer on a provider, _just_ looked at the provider side, I don't think I would understand how Pact works at all. For this reason, I think it is important to have at least one complete end-to-end example.

It might be worth creating a separate `pact-python-examples` monorepo of examples (I note that other languages have such examples). These can then go into a lot more detail on the consumer side and provider side, and might be better suited to serve as a template.

JP-Ellis:
This PR is now ready for review.

A note about merging this: It is built on top of #371 and therefore requires the latter to be merged _before_ this one can be merged. I will be rebase this PR onto the new `master` branch.

Thanks @YOU54F for all of the comments so far!

With regards to the examples not running on Cirrus, I have opted to just disable them there. The test suite should detect regressions, unlike the example only serves to illustrate the use of `pact-python`. Let me know if that's alright with you.

JP-Ellis:
In the latest changes:

- Rebase the changes onto `master` now that the hatch PR has merged (as this PR uses hatch too)
- Fixed an issue with the Message Pact example (thanks to @YOU54F for pinpointing the cause!)

This should be all good to merge 🥳

mefellows:
> The test suite should detect regressions, unlike the example only serves to illustrate the use of pact-python. Let me know if that's alright with you.

Philosophically, this makes sense to me. Especially as we look to introduce the BDD suite - making the examples be "regression" tests seems an abuse of them.

YOU54F:
>With regards to the examples not running on Cirrus, I have opted to just disable them there. The test suite should detect regressions, unlike the example only serves to illustrate the use of pact-python. Let me know if that's alright with you.

I think I would like to test them x-plat/x-arch but its not a deal-breaker for me on this PR (it was only recently introduced by myself) - assuming they aren't covering anything already covered by the existing test suite. (They almost serve at integration/regression tests for me of the main code base, which may mark its own homework with its unit test suite)

>Philosophically, this makes sense to me. Especially as we look to introduce the BDD suite - making the examples be "regression" tests seems an abuse of them.

I don't see why they can't be both. If they aren't full examples that someone can copy and paste, and they aren't regression tests, what are they then? What purpose are they trying to solve? Living documentation? Then I'd prefer they were tested against all supported platforms.

I do think the BDD suite will be a great way of unifying examples across the different codebases, but I would see them both as serving as both examples and regression suites.

examples/src/consumer.py
```diff
@@ -0,0 +1,103 @@
+"""
+Simple Consumer Implementation.
+
+This modules defines a simple
+[consumer](https://docs.pact.io/getting_started/terminology#service-consumer)
+with Pact. As Pact is a consumer-driven framework, the consumer defines the
```
YOU54F, line 6: suggestion:- which will be tested with Pact

examples/tests/test_00_consumer.py
```diff
@@ -0,0 +1,129 @@
...
+
+MOCK_URL = URL("http://localhost:8080")
+
+
+@pytest.fixture()
+def user_consumer() -> UserConsumer:
+    """
+    Returns an instance of the UserConsumer class.
+
+    As we do not want to stand up all of the consumer's dependencies, we direct
```
YOU54F, line 39: Maybe we could be more concrete here, with regards to `all` of the dependencies, and the dependencies for the UserConsumer class.

Making it clear that we are only testing our collaborater code

https://docs.pact.io/5-minute-getting-started-guide#scope-of-a-consumer-pact-test

Not the full `consumer` app.

Probably fine as is!

examples/tests/test_00_consumer.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    return UserConsumer(str(MOCK_URL))
+
+
+@pytest.fixture(scope="module")
+def pact(broker: URL, pact_dir: Path) -> Generator[Pact, Any, None]:
+    """
+    Set up Pact.
+
+    In order to test the consumer in isolation, Pact sets up a mock version of
```
YOU54F, line 52: ahh this probably covers the comment above.

examples/tests/test_00_consumer.py
```diff
@@ -0,0 +1,129 @@
...
+    the provider. This mock provider will expect to receive defined requests
+    and will respond with defined responses.
+
+    The fixture here simply defines the Consumer and Provide, and sets up the
```
YOU54F, line 56: typo : Provider

examples/tests/test_00_consumer.py
```diff
@@ -0,0 +1,129 @@
...
+    mock provider. With each test, we define the expected request and response
+    from the provider as follows:
+
+    ```python
+    pact.given("UserA exists and is not an admin") \
+        .upon_receiving("A request for UserA") \
+        .with_request("get", "/users/123") \
+        .will_respond_with(200, body=Like(expected))
```
YOU54F, line 64: is it worth being explicit about the data structure returned by our API, prior to considering writing the test, as a JSON string, rather than a dict.

Considering our provider returns the following json response

`'{"id":1,"name":"foo", "created_on": "2020-07-10 15:00:00.000"}'`

We want to consider the minimal response required for our test/consumer to pass the scenario (so we don't include extraneous information) that we don't care if the provider returns or not.

Ie if our consumer didn't need created_on for one of its scenarios (it doesn't in this case, it always requires the three fields)

examples/src/fastapi.py
```diff
@@ -0,0 +1,52 @@
+"""
+FastAPI provider example.
+
+This modules defines a simple
+[provider](https://docs.pact.io/getting_started/terminology#service-provider)
+with Pact. As Pact is a consumer-driven framework, the consumer defines the
```
YOU54F, line 6: suggestion:- which will be tested with Pact

examples/src/fastapi.py
```diff
@@ -0,0 +1,52 @@
...
+
+app = FastAPI()
+
+"""
+As this is a simple example, we'll use a simple dict to represent a database.
+This would be replaced with a real database in a real application.
+
+When testing the provider in a real application, the calls to the database
+would be mocked out to avoid the need for a real database. An example of this
+can be found in the test suite.
```
YOU54F, line 33: link to test or test filename

examples/tests/test_01_provider_fastapi.py
```diff
@@ -0,0 +1,142 @@
...
+@app.post("/_pact/provider_states")
+async def mock_pact_provider_states(state: ProviderState) -> dict[str, str | None]:
+    """
+    Define the provider state.
+
+    For Pact to be able to correctly tests compliance with the contract, the
+    internal state of the provider needs to be set up correctly. Naively, this
+    would be achieved by setting up the database with the correct data for the
+    test, but this can be slow and error-prone. Instead this is best achieved by
+    mocking the relevant calls to the database so as to avoid any side effects.
```
YOU54F, line 53: https://docs.pact.io/5-minute-getting-started-guide#testing-the-order-api-provider-project

is it worth being overt about utilising the apps db access layer (db repositories) in the above example?

examples/tests/test_01_provider_fastapi.py
```diff
@@ -0,0 +1,142 @@
...
+    examples.src.fastapi.FAKE_DB.get.return_value = None
+
+
+def mock_user_123_exists() -> None:
+    """
+    Mock the database for the user 123 exists state.
+
+    You may notice that the return value here differs from the consumer's
+    expected response. This is because the consumer's expected response is
+    guided by what the consumer users.
```
YOU54F, line 105: typo: consumer uses

examples/tests/test_01_provider_fastapi.py
```diff
@@ -0,0 +1,142 @@
...
+
+    By using consumer-driven contracts and testing the provider against the
+    consumer's contract, we can ensure that the provider is only providing what
```
YOU54F, line 108: sentence needs finishing?

we can ensure that the provider is able to safely evolve their api, whilst ensuring them meet each consumers expectations.

- Also this is a neat point for mentioning and I referred to in an earlier comment on the consumer side

examples/src/flask.py
```diff
@@ -0,0 +1,51 @@
+"""
+Flask provider example.
+
+This modules defines a simple
+[provider](https://docs.pact.io/getting_started/terminology#service-provider)
+with Pact. As Pact is a consumer-driven framework, the consumer defines the
```
YOU54F, line 6: suggestion:- which will be tested with Pact

examples/src/flask.py
```diff
@@ -0,0 +1,51 @@
...
+
+app = Flask(__name__)
+
+"""
+As this is a simple example, we'll use a simple dict to represent a database.
+This would be replaced with a real database in a real application.
+
+When testing the provider in a real application, the calls to the database
+would be mocked out to avoid the need for a real database. An example of this
+can be found in the test suite.
```
YOU54F, line 32: link to test

examples/tests/test_01_provider_flask.py
```diff
@@ -0,0 +1,135 @@
...
+def mock_user_123_exists() -> None:
+    """
+    Mock the database for the user 123 exists state.
+
+    You may notice that the return value here differs from the consumer's
+    expected response. This is because the consumer's expected response is
+    guided by what the consumer users.
+
+    By using consumer-driven contracts and testing the provider against the
+    consumer's contract, we can ensure that the provider is only providing what
```
YOU54F, line 101: missing sentence or extra word?

same comments as fast api test above :)

examples/tests/test_01_provider_flask.py
```diff
@@ -0,0 +1,135 @@
...
+    examples.src.flask.FAKE_DB.get.return_value = None
+
+
+def mock_user_123_exists() -> None:
+    """
+    Mock the database for the user 123 exists state.
+
+    You may notice that the return value here differs from the consumer's
+    expected response. This is because the consumer's expected response is
+    guided by what the consumer users.
```
YOU54F, line 98: typo: uses

examples/tests/test_02_message_consumer.py
```diff
@@ -0,0 +1,127 @@
+"""
+Test Message Pact consumer.
+
+Pact was originally designed for HTTP interactions involving a request and a
```
YOU54F, line 4: provide link to worth linking to https://docs.pact.io/getting_started/how_pact_works#non-http-testing-message-pact

and mentioning MessagePact is only responsible for the message, not the transport?

examples/tests/test_02_message_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+Pact is still consumer-driven, and the consumer defines the expected messages it
+will receive from the provider. When the provider is being verified, Pact
+ensures that the provider sends the expected messages.
+
+In this example, Pact simply ensures that the consumer is capable of processing
+the message. The consumer need not send back a message, and any sideffects of
+the message must be verified separately (such as through `assert` statements).
+
+> :warning: There is currently a bug whereby the `given` and
+`expects_to_receive` have swapped meanings. This will be addressed in a future
```
YOU54F, line 20: >have swapped meanings

maybe add 
 - compared to the reference implementation

examples/tests/test_02_message_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+
+A note on terminology, the _consumer_ for Message Pact is the system that
+receives the message, and the _provider_ is the system that sends the message.
+Pact is still consumer-driven, and the consumer defines the expected messages it
+will receive from the provider. When the provider is being verified, Pact
+ensures that the provider sends the expected messages.
+
+In this example, Pact simply ensures that the consumer is capable of processing
+the message. The consumer need not send back a message, and any sideffects of
+the message must be verified separately (such as through `assert` statements).
```
YOU54F, line 17: maybe add something to the effect of as part of usual unit testing activities.

examples/tests/test_03_message_provider.py
```diff
@@ -0,0 +1,59 @@
+"""
+Test Message Pact provider.
+
```
YOU54F, line 3: link ? 

https://docs.pact.io/getting_started/how_pact_works#non-http-testing-message-pact

examples/tests/test_03_message_provider.py
```diff
@@ -0,0 +1,59 @@
...
+Unlike the standard Pact, which is designed for HTTP interactions, the Message
+Pact is designed for non-HTTP interactions. This example demonstrates how to use
+the Message Pact to test whether a provider generates the correct messages.
```
YOU54F, line 6: >the Message Pact

MessagePact?

examples/tests/test_03_message_provider.py
```diff
@@ -0,0 +1,59 @@
...
+
+In such examples, Pact simply checks the kind of messages produced. The consumer
+need not send back a message, and any sideffects of the message must be verified
+separately.
+
+The below example verifies that the consumer makes the correct filesystem calls
+when it receives a message to read or write a file. The calls themselves are
+mocked out so as to avoid actually writing to the filesystem.
```
YOU54F, line 14: >The below example verifies that the consumer makes the correct filesystem calls
when it receives a message to read or write a file. The calls themselves are
mocked out so as to avoid actually writing to the filesystem.

Is this the right comment for this file? Should this be in the consumer side?

YOU54F (changes requested):
Great work on the documenting in the examples, just reviewed all the comments and added some feedback / suggestions / typos.

Some comments above RE: the Makefile, would love just to have commands (make test/make example - out the box as part of this PR and we can make improvements down the line) so its not in a partially working state

examples/tests/test_00_consumer.py
JP-Ellis: I've expanded on this within this docstring, and the link you suggest at the top :)

examples/tests/test_00_consumer.py
JP-Ellis: I have addressed this more explicitly when defining the `expected` response later in the test. But good call to be pointing this out explicitly.

examples/tests/test_01_provider_fastapi.py
JP-Ellis: I've added the link at the top. I'm not entirely sure what you're referring to about the apps db access layer...

examples/tests/test_01_provider_fastapi.py
JP-Ellis: Good catch, my bad!

examples/tests/test_03_message_provider.py
JP-Ellis: `MessagePact` would be the class name, but in regular English I would refer to it as "Message Pact" (same goes for the `pact-python` vs Pact Python). The space is also consistent with the docs you linked above.

examples/tests/test_03_message_provider.py
JP-Ellis: Good catch... bad copy-pasta on my end 😆

examples/tests/test_01_provider_fastapi.py
YOU54F: Ahh so 

we have a fake db here in the provider code

https://github.com/pact-foundation/pact-python/blob/f9d2024b19237f1005bee592691ecf237270093e/examples/src/fastapi.py#L35C1-L35C8

and we mock it out on the provider test, with a magic mock

https://github.com/pact-foundation/pact-python/blob/f9d2024b19237f1005bee592691ecf237270093e/examples/tests/test_01_provider_fastapi.py#L113

i suppose in your prod code, you might have an abstraction layer which say says a database with dependency injection, so it could easy use a fake db or a real db to fulfil the db access functions required by the application code (so the main logic of the code, and the tests are agnostic to the fake the database is _real_ or not.

YOU54F: this is fine as is anyway, just thinking aloud :)

JP-Ellis:
Thanks @YOU54F for the amazing review!

I've addressed the inline comments and have fixed the Makefile. I've also made sure the examples run on Python 3.8.

I still think that the e2e examples need not be run in the CI for all platforms, primarily because it relies on pulling and setting up a Docker image which slows things down a fair bit.

examples/tests/test_01_provider_fastapi.py
JP-Ellis: Your production code would have just the production code. This may be abstracted through some layer, or provided by a direct dependency. The idea though is that the test should not change the production code, and instead the test suite does some magic mocking and intercepts the actual calls and returns what is expected.

Do you think it is worth elaborating in the docs?

YOU54F (approved):
Awesome, thanks for the update, looks great

PR #388, opened 2023-09-15
fix: support httpx >= 0.23.3 for python >= 3.7

codenaz:
When installing this library in a project with httpx `0.24.1`, it throws the follwing error

```
Because no versions of pact-python match >2.0.1,<3.0.0
 and pact-python (2.0.1) depends on httpx (0.23.3), pact-python (>=2.0.1,<3.0.0) requires httpx (0.23.3).
So, because project depends on both httpx (^0.24.1) and pact-python (^2.0.1), version solving failed.
```

This pull request adds support for higher httpx versions


JP-Ellis:
Thanks for the PR! Much appreciated.

There's an active PR which is upgrading all of these dependencies (and much more) in #371. If you are happy to wait a little bit, and we should be release a `v2.1` soon :) It is currently waiting for further review as it is a rather substantial change.

codenaz:
> Thanks for the PR! Much appreciated.
> 
> There's an active PR which is upgrading all of these dependencies (and much more) in #371. If you are happy to wait a little bit, and we should be release a `v2.1` soon :) It is currently waiting for further review as it is a rather substantial change.

Awesome! I will happily give it a look and perhaps give some feedback. Excited this is happening

YOU54F:
Thanks @codenaz for offering the change, and feedback is hugely welcome and greatly appreciated on any of the pull requests.

PR #394, opened 2023-09-25
Release: v2.1.0

JP-Ellis:
Update pact standalone and bump version to get v2.1.0 out.

@YOU54F Not sure what your previous release process has been like :)


YOU54F:
i think someone is up with the changelog. we’ve got two files now?

it’s good to release small and release often, it’s all automated. the releasing document at the base of the repo should be up to date

once a release workflow is triggered, a maintainer will need to approve ( the user triggering the release can bypass that check with a note )

JP-Ellis:
Yeah, I was a little confused with the changelogs too. I noticed that commitizen is being used, but it produced a completely different changelog to the existing one. Also, the existing one appears to be a git log as opposed to an actual changelog (and end-users of Pact Python don't really need to know the chore commits in general, so the changelog need not include them).

YOU54F:
Ahhh, 

so the `script/release_prep.sh` is run for releases, which generates a changelog from the git log history, rather than with commitizen

https://github.com/pact-foundation/pact-python/blob/7c60dd5e0979907828b27ff7b9bb5069eec397c8/script/release_prep.sh#L21C3-L28C50

It would be good to update how you are generating the version number?  and maybe update the release_prep.sh script.

PS, would you prefer releases to come via Pull reqs - currently they are triggered from HEAD 

https://github.com/pact-foundation/pact-python/blob/master/RELEASING.md#preparing-the-release

which pushes back the updated changelog via the current process

JP-Ellis:
Fixed the doubling-up of changelogs, and created #400 to discuss updating the changelog format :) Let me know what you think. I'll push the `v2.1.0` tag once this branch is merged.

YOU54F (approved):

Issue #384, opened 2023-09-13
Create a Python wheel to bundle the Pact-Rust shared libs


mefellows:
- [x] Extend the `hatch` build process to download the Pact reference library
- [x] Extend the `hatch` build process to build an FFI module
- [x] Implement a rudimentary test to ensure it works.

## Original Description

Python has very good support for platform specific python packages called wheels. See [wheels](https://realpython.com/python-wheels)

This story is to create a new project that can create a Python wheel that bundles the Pact-Rust libs. This would need a CI build that can create a wheel for the major OS versions and architectures and a release process to upload to PyPI.

Note that using a wheel means it will not work with Alpine Linux.

> *Caution*: Be advised that [PyPI wheels don’t work on Alpine Linux](https://pythonspeed.com/articles/alpine-docker-python) (or [BusyBox](https://hub.docker.com/_/busybox/)). This is because Alpine uses `musl` in place of the standard `glibc`. The `musl libc` library bills itself as “a new `libc` striving to be fast, simple, lightweight, free, and correct.” Unfortunately, when it comes to wheels, `glibc` it is not

Issue #242, opened 2021-07-15
Installer


elliottmurray:
Requires the FFI artifacts to be installed. Pact-Python setup  command to install the required libs into the correct location.

Done when the Pact Rust CLI Verify command can  be executed correctly after installing Pact-Python

YOU54F:
Completed for all available architectures and platforms in #357 following on from the work done in #245

PR #395, opened 2023-09-26
FFI Groundwork

JP-Ellis:
This PR sets the groundwork to use the FFI. It ensures that the FFI is built during the wheel creation process thereby avoiding runtime slowdowns.

I have made sure that the commits are self contained for the review process. The commit messages are:

-   **chore(ci): update build targets**
    
    As the upstream Pact reference library has a different set of targets,
    the build targets for this library have been updated to match. The most
    significant change is the dropping is 32-bit architectures altogether.
    This also adds a `musllinux` target (which was previously not
    supported).

-   **feat(v3): add v3.ffi module**
    
    This module provides a Python interface to the Pact library written in
    Rust. For this first commit, only the `pactffi_version()` function is
    implemented and tested.
    
    In the transition to v3, the new codebase will be located within the
    `v3` submodule. This will allow the v2 code to remain in place for
    backwards compatibility, and will allow the v3 code to be tested
    independently of the v2 code.
    
    Once the v3 code is complete, the existing v2 code will be scoped to a
    new `v2` submodule, and the v3 code will be moved to the root of the
    repository.

-   **chore(tests): add ruff.toml for tests directory**
    
    Adjust the lint rules that apply to tests to be more permissive.
    Specifically to allow the use of `assert` statements.

-   **style!: refactor constants**
    
    With the possibility of building wheels against systems for which no
    Ruby executables exist, the constants module has been refactored to
    allow for the use of system installed Pact executables.
    
    This also introduces the `USE_SYSTEM_PACT` environment variable which
    can be used to force the use of system installed Pact executables.
    
    In doing these changes, the module has been refactored to avoid
    redundancies, and avoids the complexities of Windows executables
    extensions by using the `shutil.which` function.
    
    The test was refactored accordingly to test the constants instead of
    the functions.
    
    BREAKING CHANGE: The public functions within the constants module have
    been removed. If you previously used them, please make use of the
    constants. For example, instead of `pact.constants.broker_client_exe()`
    use `pact.constants.BROKER_CLIENT_PATH` instead.
    
    BREAKING CHANGE: It is possible to use the system installed Pact
    executables by setting `USE_SYSTEM_PACT` to `True` or `Yes`.

-   **chore(build): update packaging to build ffi**
    
    As we will transition to using the Rust Pact library, we need to
    download it as part of the build process. This commit adds a step to the
    build process to download the library and extract it to the correct
    location and then build the Python bindings.
    
    As the Rust library is available for more platforms than the Ruby
    executables, failing to find the Ruby executables is no longer a fatal
    error and will instead be raised as a warning during the build process.
    
    So small adjustments to the build script were made to accommodate this
    change.

A note about the CI/CD: Python 3.12 drops support for `distutils` (see [PEP 632](https://peps.python.org/pep-0632/)). The latest stable version of `cffi` at the time of creating this PR is [`1.15`](https://cffi.readthedocs.io/en/release-1.15/) which still relies on `distutils`. This is resolved in the [`1.16` release candidate](https://cffi.readthedocs.io/en/release-1.16/whatsnew.html#v1-16-0rc1) and there won't be a long term issue.

Resolves: #384
Resolves: #242

### DIFF
--- a/hatch_build.py
+++ b/hatch_build.py
@@ -1,33 +1,56 @@
-"""Hatchling build hook for Pact binary download."""
+"""
+Hatchling build hook for binary downloads.
+
+Pact Python is built on top of the Ruby Pact binaries and the Rust Pact library.
+This build script downloads the binaries and library for the current platform
+and installs them in the `pact` directory under `/bin` and `/lib`.
+
+The version of the binaries and library can be controlled with the
+`PACT_BIN_VERSION` and `PACT_LIB_VERSION` environment variables. If these are
+not set, a pinned version will be used instead.
+"""
 
 from __future__ import annotations
 
+import gzip
 import os
 import shutil
-import typing
+import tarfile
+import tempfile
+import warnings
+import zipfile
 from pathlib import Path
 from typing import Any, Dict
 
+import cffi
+import requests
 from hatchling.builders.hooks.plugin.interface import BuildHookInterface
 from packaging.tags import sys_tags
 
 ROOT_DIR = Path(__file__).parent.resolve()
-PACT_VERSION = "2.0.7"
-PACT_URL = "https://github.com/pact-foundation/pact-ruby-standalone/releases/download/v{version}/pact-{version}-{os}-{machine}.{ext}"
-PACT_DISTRIBUTIONS: list[tuple[str, str, str]] = [
-    ("linux", "arm64", "tar.gz"),
-    ("linux", "x86_64", "tar.gz"),
-    ("osx", "arm64", "tar.gz"),
-    ("osx", "x86_64", "tar.gz"),
-    ("windows", "x86", "zip"),
-    ("windows", "x86_64", "zip"),
-]
+
+PACT_BIN_VERSION = os.getenv("PACT_BIN_VERSION", "2.0.7")
+PACT_BIN_URL = "https://github.com/pact-foundation/pact-ruby-standalone/releases/download/v{version}/pact-{version}-{os}-{machine}.{ext}"
+
+PACT_LIB_VERSION = os.getenv("PACT_LIB_VERSION", "0.4.9")
+PACT_LIB_URL = "https://github.com/pact-foundation/pact-reference/releases/download/libpact_ffi-v{version}/{prefix}pact_ffi-{os}-{machine}.{ext}"
 
 
-class PactBuildHook(BuildHookInterface):
+class PactBuildHook(BuildHookInterface[Any]):
     """Custom hook to download Pact binaries."""
 
     PLUGIN_NAME = "custom"
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:  # noqa: ANN401
+        """
+        Initialize the build hook.
+
+        For this hook, we additionally define the lib extension based on the
+        current platform.
+        """
+        super().__init__(*args, **kwargs)
+        self.tmpdir = Path(tempfile.TemporaryDirectory().name)
+        self.tmpdir.mkdir(parents=True, exist_ok=True)
 
     def clean(self, versions: list[str]) -> None:  # noqa: ARG002
         """Clean up any files created by the build hook."""
@@ -43,10 +66,10 @@ class PactBuildHook(BuildHookInterface):
         build_data["infer_tag"] = True
         build_data["pure_python"] = False
 
-        pact_version = os.getenv("PACT_VERSION", PACT_VERSION)
-        self.install_pact_binaries(pact_version)
+        self.pact_bin_install(PACT_BIN_VERSION)
+        self.pact_lib_install(PACT_LIB_VERSION)
 
-    def install_pact_binaries(self, version: str) -> None:  # noqa: PLR0912
+    def pact_bin_install(self, version: str) -> None:
         """
         Install the Pact standalone binaries.
 
@@ -54,10 +77,30 @@ class PactBuildHook(BuildHookInterface):
         the current operating system is determined automatically.
 
         Args:
-            version: The Pact version to install. Defaults to the value in
-            `PACT_VERSION`.
+            version: The Pact version to install.
         """
-        platform = typing.cast(str, next(sys_tags()).platform)
+        url = self._pact_bin_url(version)
+        if url:
+            artifact = self._download(url)
+            self._pact_bin_extract(artifact)
+
+    def _pact_bin_url(self, version: str) -> str | None:  # noqa: PLR0911
+        """
+        Generate the download URL for the Pact binaries.
+
+        Generate the download URL for the Pact binaries based on the current
+        platform and specified version. This function mainly contains a lot of
+        matching logic to determine the correct URL to use, due to the
+        inconsistencies in naming conventions between ecosystems.
+
+        Args:
+            version: The upstream Pact version.
+
+        Returns:
+            The URL to download the Pact binaries from, or None if the current
+            platform is not supported.
+        """
+        platform = next(sys_tags()).platform
 
         if platform.startswith("macosx"):
             os = "osx"
@@ -67,10 +110,16 @@ class PactBuildHook(BuildHookInterface):
                 machine = "x86_64"
             else:
                 msg = f"Unknown macOS machine {platform}"
-                raise ValueError(msg)
-            url = PACT_URL.format(version=version, os=os, machine=machine, ext="tar.gz")
+                warnings.warn(msg, RuntimeWarning, stacklevel=2)
+                return None
+            return PACT_BIN_URL.format(
+                version=version,
+                os=os,
+                machine=machine,
+                ext="tar.gz",
+            )
 
-        elif platform.startswith("win"):
+        if platform.startswith("win"):
             os = "windows"
 
             if platform.endswith("amd64"):
@@ -79,11 +128,16 @@ class PactBuildHook(BuildHookInterface):
                 machine = "x86"
             else:
                 msg = f"Unknown Windows machine {platform}"
-                raise ValueError(msg)
+                warnings.warn(msg, RuntimeWarning, stacklevel=2)
+                return None
+            return PACT_BIN_URL.format(
+                version=version,
+                os=os,
+                machine=machine,
+                ext="zip",
+            )
 
-            url = PACT_URL.format(version=version, os=os, machine=machine, ext="zip")
-
-        elif "linux" in platform:
+        if "linux" in platform and "musl" not in platform:
             os = "linux"
             if platform.endswith("x86_64"):
                 machine = "x86_64"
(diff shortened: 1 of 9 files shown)


mefellows:
This is great!

> BREAKING CHANGE: It is possible to use the system installed Pact
 executables by setting USE_SYSTEM_PACT to True or Yes.

I didn't quite follow this (what was the previous behaviour and what is the new behaviour). Would this be obvious to users what it means? 

> Once the v3 code is complete, the existing v2 code will be scoped to a
new v2 submodule, and the v3 code will be moved to the root of the
repository.

Just so I'm clear. Given that the change in this PR is a breaking change (and thus requires a major version bump), this proposal would then require another major version bump when that package rework happens? I think that's probably fine, and preferably to having a long lived fork.

One question arises about major version support lifetime, and if bugs will be addressed/backported to them - will you create release branches for the previous major versions?

pact/v3/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""
+Pact Python V3.
+
+The next major release of Pact Python will make use of the Pact reference
+library written in Rust. This will allow us to support all of the features of
+Pact, and bring the Python library in line with the other Pact libraries.
+
+The migration will happen in stages, and this module will be used to provide
```
mefellows, line 8: Nice!

pact/v3/ffi.py
```diff
@@ -0,0 +1,18 @@
+"""
```
mefellows, line 1: Question around the library packaging/design. The FFI is agnostic to the version of any Python API interface, so you may want to consider separating it from a `v3` or `v4` API. I believe the folder is more about representing a major version bump (rather than pact specification versions), in which case, this is probably fine.

I assume the ffi functions will also be kept internal to Pact Python and not part of the public API, and therefore allow the flexibility of moving about/refactoring as needed.

mefellows (approved):
I have a few questions, but I think the general gist of this is good.

pact/v3/ffi.py
JP-Ellis: So first your question around the `v3` naming. I noticed in the existing PR that the classes have been renamed to include a `V3` suffix (e.g., `ConsumerV3`, `PactV3`, etc.). I personally think that using the same class name, but under a different naming scope will better suit Python. This gives the end user a fair bit of flexibility, with all of the below being possibility:

```python
### What I would likely use
from pact import Consumer
import pact.v3 as pv3
Consumer(...)
pv3.Consumer(...)

### but one could also rename on import
from pact import Consumer as ConsumerV2
from pact.v3 import Consumer as ConsumerV3
ConsumerV2(...)
ConsumerV3(...)

### Or if someone wants to just use v3
from pact.v3 import Consumer
```

And the version naming here is specific to the Pact Python version (currently at 2, and it would migrate to 3 when this change is fully merged). It is not a reference to the Pact specification. I assume that the Pact reference library supports both v3 and v4 of the Pact specification.

---

As to your 2nd question about around public/internal functions (Python has no concept of private unfortunately). By convention, a prefix `_` indicates that a module/function is private.

The `pact.v3._ffi` module contains the binary extension which is automatically generated based on the upstream library. It is generally quite difficult to use these, and requires a lot of casting between Python and C types.

The `pact.v3.ffi` will contain the Python-wrapped functions from the `_ffi` modules. I doubt many people would use these functions, but the naming technically means they are public. The reasoning here is that I am assuming these would be stable in as far as the upstream Pact library is stable. If an upstream function is experimental or deprecated, I can quite easily create the appropriate warnings.

JP-Ellis:
> This is great!

Thanks 😁 

> > BREAKING CHANGE: It is possible to use the system installed Pact
> > executables by setting USE_SYSTEM_PACT to True or Yes.
> 
> I didn't quite follow this (what was the previous behaviour and what is the new behaviour). Would this be obvious to users what it means?

The current implementation bundles the Ruby executables and explicitly calls them. This was fine as long as we only distributed wheels to platforms for which Ruby executables exist; however, the platforms supported by the Rust library are different.

This would not be a breaking change for the vast majority of people, and most users would not notice this. So while yes it is technically a breaking change, the impact on end users is rather negligible.


> Just so I'm clear. Given that the change in this PR is a breaking change (and thus requires a major version bump), this proposal would then require another major version bump when that package rework happens? I think that's probably fine, and preferably to having a long lived fork.
> 
> One question arises about major version support lifetime, and if bugs will be addressed/backported to them - will you create release branches for the previous major versions?

This breaking change is the only one I think will be necessary while working on the `pact.v3` module. As the impact is very very minimal (and the remedy is simple), I am thinking we might be able to get away with a minor version bump only. I still would want to make sure this is thoroughly explained in the release notes in case anyone has any issues.

JP-Ellis:
Just fixed a typo preventing the linux aarch64 library from being downloaded.

@YOU54F, whenever you have time, I'd love to hear your thoughts on this (and if you're too busy, happy to proceed with Matt's approval).

pact/constants.py
```diff
@@ -1,38 +1,60 @@
...
+
+_USE_SYSTEM_PACT = os.getenv("USE_SYSTEM_PACT", "").upper() in ("TRUE", "YES")
+_BIN_DIR = Path(__file__).parent.resolve() / "bin"
+
+
+def _find_executable(executable: str) -> str:
+    """
+    Find the path to an executable.
+
+    This inspects the environment variable `USE_SYSTEM_PACT` to determine
```
YOU54F, line 29: does this have to be available on the path?

could the user have the possibly to provider their own path, if they don't want to or struggle to have it available on their path?

I think a `PACT_XYZ` env var might be more idiomatic with env vars set in other languages, the pact broker and the pact cli 

say PACT_USE_SYSTEM_LIBS

you could also then have potentially a CUSTOM_LIBS location.

I think @mefellows would love in a config living in a users `$HOME/.pact` directory and the plugins are now installed into their.

They take a PACT_PLUGIN_DIR location.

As this is for the CLI tools? it could be PACT_CLI_DIR

hatch_build.py
```diff
@@ -67,81 +110,314 @@ def install_pact_binaries(self, version: str) -> None:  # noqa: PLR0912
...
+        warnings.warn(msg, RuntimeWarning, stacklevel=2)
+        return None
+
+    def _pact_bin_extract(self, artifact: Path) -> None:
+        """
+        Extract the Pact binaries.
+
+        The upstream distributables contain a lot of files which are not needed
+        for this library. This function ensures that only the files in
+        `pact/bin` are extracted to avoid unnecessary bloat.
```
YOU54F, line 167: the files in pact/bin are only wrapper scripts which call the files in the lib directory, which contains the ruby runtime and the actual pact exectuables.

https://github.com/pact-foundation/pact-ruby-standalone/blob/7d6467e942884cf88f171de34bd1a6b5dbe94100/packaging/pact-broker.sh#L29

Screenshot 2023-10-05 at 19 44 58

The unit tests are just checking for the wrapper scripts existence.

pact/constants.py
JP-Ellis: Very good points!

- I will rename the system variable to be `PACT_...`

- The `USE_SYSTEM_PACT` described above was more specifically for the Ruby executables. Upon reflection, I really should have:
  - `PACT_USE_SYSTEM_BINS` for the CLI tools
  - `PACT_USE_SYSTEM_LIBS` for the libraries. Though I'm not entirely sure how/if that will work.

- Specifically regarding `PACT_USE_SYSTEM_LIBS`, at the moment the build script pulls down the static libraries and then creates shared library against the system libraries. Specifically, library created contains the Pact lib in its entirety (and therefore doesn't rely on a `libpact_ffi` being installed in `/usr/lib`).

- The above variables are simply booleans and use the system's default resolution for executables and libraries. I generally don't really like hijacking the way specifying direct paths, as there is a strong Unix convention already of updating `PATH`. For example:

  ```sh
  PACT_USE_SYSTEM_BINS=True PATH=/path/to/pact:$PATH pytest ...
  ```

  This is particularly important, in my opinion, for the Pact libraries too as the loading of libraries is done much less explicitly than the calling of executables. By setting `LD_LIBRARY_PATH` as appropriate, there is much less risk of accidentally loading the wrong library.

- Regarding the Pact plugin's, at the moment, this is not supported at all, but I agree it will need to be done eventually. I don't think it is within the scope of this PR for now and I'll get to it once I implement plugins in the library.

hatch_build.py
JP-Ellis: For some reason, I had the idea that Ruby managed to combine the libs into a single executable... That is my mistake!

I'm rather worried now that the unit tests don't actually test the library properly.

JP-Ellis:
Thanks for all of the testing! 

> ## Hatch - extra test args, or selective test running
> I also updated in the my local testing, the hatch test command
> 
> `test = "pytest {args:--cov-config=pyproject.toml --cov=pact --cov=tests tests}"`
> 
> so I could pick up args for running with `-rP` now with `hatch run test tests -rP` or a single test with `hatch run test/test_ffi.py -rP` (i wanted to see the std out printing of the lib just for completeness in my world xD) Maybe you have a more elegant way! Basically I want to pass extra args, or run a subset of the tests

For some reason, I had it in my mind that I have already done this... but clearly not 😅. This was definitely the intention and let me fix that up now!

> ## Dockerfiles
> I've noted the need to update the Dockerfiles as they still use tox, I updated them locally as part of my testing
> 
> I'll drop a separate PR to update those
> 
> I will also
> 
> * move the ones at the root of the repo, into the `/docker` folder.
> * remove the ubuntu one, it isn't necessary as we have a debian flavour

I was considering migrating the various Dockerfiles to a Dev Container. I have created an issue to track this at #407.

pact/v3/ffi.py
mefellows: Thanks. That clarifies it for me.

I agree, having versions in the types themselves is distasteful - I regret having that in the JS library.

> I assume that the Pact reference library supports both v3 and v4 of the Pact specification.

That's correct. However as you'll see, there are different capabilities that come with each version of the spec (particularly on the consumer side). As long as your consumer API is able to constrain the options based on which version of the spec they intend to use, you should be OK.

For example, you can't say "I want to use V2 of the spec" and then go on to define a synchronous message interaction (introduced in V4).  

If you'd like to see that expressed as types (and using the type-state pattern), [this](https://github.com/pact-foundation/pact-js/blob/master/src/v4/http/types.ts) might be helpful. Obviously, the [compatibility suite](https://github.com/pact-foundation/pact-compatibility-suite/) would be a good reference also.

hatch_build.py
JP-Ellis: This has been fixed in 99f7657

pact/v3/ffi.py
JP-Ellis: As I work through to implementing the entire compatibility suite, I'll make sure that the correct Pact spec is available, and can be constrained :)

JP-Ellis:
I have updated this PR to:

- Incorporate changes as suggested by @YOU54F. Specifically:
  - The `USE_SYSTEM_PACT` environment variable has been changed to `PACT_USE_SYSTEM_BINS`. Also updated the commit message and the `BREAKING CHANGE` docs.
  - I have fixed the exclusion of the `lib/` directory.
- Be rebased on top of the more recent changes in `master`, resolving the merge conflict.

YOU54F:
Great thanks dude, will take a look this morning

YOU54F (approved):
Tested this out locally and it was all good, nice work, thanks for getting the ruby runtime working again, and great to see alpine officially supported for the ffi.

🙌🏾

JP-Ellis:
Did a rebase to resolve a merge conflict. Checked the failing tests, and they are purely related to the release of Python 3.12 (and can be fixed at a later date, through #405).

PR #403, opened 2023-10-02
FFI Wrapper

JP-Ellis:
## :memo: Summary

Implement a `pact.v3.ffi` module which exposes functions to cover all of the functions exposed by the Pact library.

Due to the number of functions, most of these are simply:

```py
def foo(...):
    """
    {description from pact.h}
    """
    raise NotImplementedError
```

and they will be filled in as the implementation proceeds.

Most of the code in this PR was automatically created, and then modified slightly to ensure conformance with Ruff's lints.

## :hammer: Test Plan

As the rest of the `pact.v3` module gets populated, more of these functions will get tested. At this stage, only `version` is defined and tested.

All tests for the new `pact.v3` modules will be stored under the corresponding `tests/v3/` directory.

## :link: Related issues/PRs

- Resolves: #399 

  Turns out that implementing a wrapper for _all_ functions was a bit ambitious given the number of functions.

- #395 

  That PR must be merged _before_ this PR.


pact/v3/ffi.py
YOU54F, line 1: so this is just v3 interface for pact-python (which is currently on v2 for pact-python and supports up to v2 of the pact-spec), not any relation to the ffi's pact specification capabilities. I'm okay with that, but I just know how confusing it was in Pact-Net.

Pact-Net v4 came out, which was the first to support Pact V3 :D 

so yeah as long as we are clear about Pact Specification support, for any particular version of Pact-Python in a table, that will suffice :)

YOU54F:
Hey had a quick look at this, should this have master merged into it now? or master merged into the ffi-groundwork PR? Would make it a bit easier to review in a ui diff I think.

pact/v3/ffi.py
JP-Ellis: I think we have things confused... The `v3` here refers to the Pact Python version. The aim here is that `python-python@v3` will support all current Pact specifications (V1 through to V4).

YOU54F: spot on! crystal clear for me now.

JP-Ellis:
Yeah sorry for the review being difficult. It's built on top of the FFI groundwork, hence why it has a lot of extra commits. Only the last commit, 0afc711, is new.

YOU54F:
no need to apologise, wasn't sure if it was because of some git merge mastery that you have lined up.

i ended up merging master in locally in the previous #395 as I wanted to run the examples, after running and checking the PR as it was

JP-Ellis:
I have rebased this PR on top of the `feat/initial-ffi` branch (as it was itself rebased). I've also adjusted to target branch for this PR to be the latter to help single out the changes include here.

YOU54F (approved):
good stuff!

JP-Ellis:
Rebased on top of `master` to ensure it can merge. Only failing tests are related to the recent release of Python 3.12.

Issue #382, opened 2023-09-13
Add support and example for a HTTP consumer test using Rust engine


mefellows:
“Add support” means adding DSL methods that call the underlying Rust functionality.

“Add example” means create an example Python project that showcases using the functionality.

Issue #376, opened 2023-09-13
Add support and example of a HTTP consumer test using a plugin


mefellows:
“Add support” means adding DSL methods that call the underlying Rust functionality.

“Add example” means create an example Python project that showcases using the functionality.

JP-Ellis:
I think this can actually be closed in favour of #1092 as that uses an example/

PR #410, opened 2023-10-06
Pact Consumer

JP-Ellis:
## :memo: Summary

Implements the core `Pact` class, along with a number of other classes required for this to all work.

## ~:rotating_light: Breaking Changes~

No changes to the existing library, but introduces significant changes to `pact.v3`.

## :hammer: Test Plan

Unit tests have been included as part of this PR.

## :link: Related issues/PRs

- Resolves: #382

## :eyes: Notes for Reviewers

As this PR is rather substantial, I have split some of the less consequential commits from the major commit (`feat(v3): implement pact class`). While the PR as a whole can be reviewed, I would recommend reviewing each commit individually.

### DIFF
--- a/pact/v3/ffi.py
+++ b/pact/v3/ffi.py
@@ -16,15 +16,56 @@ This modules should provide the following only:
 -   Basic Enum classes
 -   Simple wrappers around functions, including the casting of input and output
     values between the high level Python types and the low level C types.
+-   Simple wrappers around some of the low-level types. Specifically designed to
+    automatically handle the freeing of memory when the Python object is
+    destroyed.
 
 These low-level functions may then be combined into higher level classes and
-modules.
+modules. Ideally, all code outside of this module should be written in pure
+Python and not worry about allocating or freeing memory.
 
 During initial implementation, a lot of these functions will simply raise a
 `NotImplementedError`.
 
 For those unfamiliar with CFFI, please make sure to read the [CFFI
 documentation](https://cffi.readthedocs.io/en/latest/using.html).
+
+### Handles
+
+The Rust library exposes a number of handles to internal data structures. This
+is done to avoid exposing the internal implementation details of the library to
+users of the library, and avoid unnecessarily casting to and from possibly
+complicated structs.
+
+In the Rust library, the handles are thin wrappers around integers, and
+unfortunately the CFFI interface sees this and automatically unwraps them,
+exposing the underlying integer. As a result, we must re-wrap the integer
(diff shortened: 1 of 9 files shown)


pact/v3/ffi.py
```diff
@@ -16,26 +16,79 @@
...
 documentation](https://cffi.readthedocs.io/en/latest/using.html).
+
+### Handles
+
+The Rust library exposes a number of handles to internal data structures. This
+is done to avoid exposing the internal implementation details of the library to
+users of the library, and avoid unnecessarily casting to and from possibly
+complicated structs.
+
+In the Rust library, the handles are thin wrappers around integers, and
```
rholshausen, line 40: This was done to make it easier for Go and JS, but if it is a problem, we can change it to be a pointer type

rholshausen (approved):
LGTM

pact/v3/ffi.py
JP-Ellis: It's not at all an issue. It's just that a `struct { int }` has the same in-memory representation as a bare `int`. I added the thin wrappers for two reasons:

- Obfuscating the `int`. After all, it doesn't make sense to 'add' or 'subtract' handles (unlike integers). It also makes it explicit that the handles to a Pact or an Interaction are different.
- This makes it easy to implement the `__del__` method to ensure that memory is cleaned up when Python's garbage control executes.

YOU54F: Great explanation @JP-Ellis 👍🏾 thank you!

YOU54F (approved):
awesome!!!

YOU54F:
>🔗 Related issues/PRs
Resolves: https://github.com/pact-foundation/pact-python/issues/376

Should this be Add support and example of a HTTP consumer test without the plugin bit. I couldn't see a plugin being loaded in any of the unit tests.

The methods are exposed, but there isn't an example unit test as per linked ticket (that i've seen)

YOU54F:
also added `g++` to the linux image deps to support a couple of python versions that were failing to build

JP-Ellis:
> > 🔗 Related issues/PRs
> > Resolves: https://github.com/pact-foundation/pact-python/issues/376
> 
> Should this be Add support and example of a HTTP consumer test without the plugin bit. I couldn't see a plugin being loaded in any of the unit tests.
> 
> The methods are exposed, but there isn't an example unit test as per linked ticket (that i've seen)

I linked to the wrong underlying issue 😅. Fixed.

As with the related PRs, I had to rebase this before merging it.

PR #414, opened 2023-10-10
docs: setup mkdocs

JP-Ellis:
## :memo: Summary

MkDocs is a popular tool in the Python ecosystem to generate documentation. It has support for parsing Python docstrings in order to generate API documentation quite easily.

The docs can be seen live at:

> [`pact-foundation.github.io/pact-python/`](https://pact-foundation.github.io/pact-python)

Note that a lot of the 'legacy' codebase will not render correctly. The sections which should be looked at are:

- [`pact.v3`](https://pact-foundation.github.io/pact-python/pact/v3/)
- The [examples](https://pact-foundation.github.io/pact-python/examples/) docs

> [!IMPORTANT]  
> Before this PR is merged, the publication step of the Docs GitHub Action must be modified so that it only runs on `master`.

## :fire: Motivation

Work for this was created during my own development to see the API from an end-users' perspective.


JP-Ellis:
The work is finally ready for review, and can be viewed online at:

> https://pact-foundation.github.io/pact-python/


> [!IMPORTANT]  
> Before this PR is merged, the publication step of the Docs GitHub Action must be modified so that it only runs on `master`.

@YOU54F and @mefellows, I would be keen for your review on the docs before it gets merged as this is a rather substantial change, and I want to make sure there are no unforeseen repercussions.

JP-Ellis:
Closes #416

PR #424, opened 2023-10-19
Feat/interaction implementation

### LINKED ISSUE #419: Implement all FFI methods relating to `Interaction`
## Description

- [x] Implement all missing methods for the `Interaction` class. The scope of 'all' here relates to the methods exposed by the FFI which typically either: construct a Interaction (returning an `InteractionHandle`); or, take an `InteractionHandle` as an argument and modify the underlying Pact
- [x] Ensure all new methods are systematically unit-tested.

## Has this been requested on Canny?

_No response_

## Motivation

Initial work implementing the `Interaction` class was done as part of #410; however, the scope of the PR was limited to a MWE to ease with the review and ensure that the more substantive/foundational parts can be reviewed and the finer implementation details can kept separate.

## Have you tried building it?

Continuation of prior work

## Self-service

- [X] I'd be willing to contribute this feature to Pact Python myself.

JP-Ellis:
## :memo: Summary

This adds implementation for all FFI functions which act on an `InteractionHandle`. Most of these map in Python to a method of the `Interaction` class.

As the InteractionHandle can point to a HTTP, Sync Message or Async Message Interaction, the initial `Interaction` class has been converted into an abstract class, and then subclassed into three new concrete classes. This will help end-users specifically when it comes to auto-completions within the IDE.

A couple of minor changes have also been included in this PR in separate commits:

-   **chore(test): automatically generated xml coverage**
    
    This can be used by other tools to generate coverage reports, or to annotate code within the IDE.
    
-   **fix(v3): unconventional __repr__ implementation**
    
    The __repr__ implementation should produce one of:
    
    1. A valid Python string which could be passed to `eval()` to generate the same object.
    2. A string of the form `<{class_name}: {info}>`.
    
    A few of the implementations did not quite adhere to this convention and have been fixed.

## ~:rotating_light: Breaking Changes~

Only impacting v3 submodule, which makes not stability guarantees.

## :fire: Motivation

Part of the development of Pact Python version 3

## :hammer: Test Plan

Unit tests have been included, with a couple of caveats:

- I was unable to get the multipart matching to work. While Pact is set up to expect a multipart payload, I have been unable to get the `aiohttp` and Pact libraries to generate mutually compatible payloads. As this is one of the lesser used feature, I think it is reasonable for have this merged for now and add a ticket to the backlog.
- The `SyncMessageInteraction` and `AsyncMessageInteraction` are implemented, but only insofar as to test their instantiation. They are not tested any further.

## :link: Related issues/PRs

- Resolves #419


rholshausen (approved):
LGTM

Issue #420, opened 2023-10-11
Implement HTTP consumer compatibility suite


JP-Ellis:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/pact-foundation/pact-python/blob/main/CONTRIBUTING.md#issues).

### Description

Implement the following consumer compatibility suites:

- [x] #473
- [x] #474
- [x] #475
- [x] #476

### Motivation

To ensure compliance and consistency across Pact implementations, the Pact Reference repository has created a compatibility suite. As part of the initial work

### Have you tried building it?

Not yet.

### Self-service

- [X] I'd be willing to contribute this feature to Pact Python myself.

JP-Ellis:
Closed as all sub-tasks are done 🥳

PR #468, opened 2023-11-12
Consumer Compatibility Suite (V1)

JP-Ellis:
## :memo: Summary

This implements the first (of many) compatibility suite tests: `V1/http_consumer.feature`.

This PR focuses primarily on the compatibility suite, but also includes a few changes to the `v3` module in order to implement features required for the compatibility suite, or bugs identified as a result of the compatibility suite.

A couple of scenarios from the feature file have been disabled for now, pending on a resolution from upstream.

### Note to the reviewer

As this PR includes a few changes to other parts of the codebase, it might be best to review the PR on a per-commit basis.

None related to the existing `v2` module

## :hammer: Test Plan

Compatibility suite is the test plan.

## :link: Related issues/PRs

- Part of #420 
- Resolves: #473


rholshausen:
LGTM

mefellows (approved):
Nice work! I mostly just browsed through it rather than fine tooth comb given the volume. The regex and other syntax for the parameterization is obscure/interesting, but it looks to have done the job!

JP-Ellis:
It should is pretty standard Python regex, with the main uncommon feature being the naming of groups using `(?P<name>...)`. This is required to ensure the group can be referenced by name 🙂

mefellows:
Oh, I just meant it's kind of funky to read it in the descriptions that way. The `negation` thing was also cool.

YOU54F:
Might be worth adding in the docs that to run these, one would have to pull in the submodules

`git submodule update --init`

YOU54F (approved):

YOU54F:
cirrus-ci looks like they are failing due to not checking out submodules 👍🏾

PR #539, opened 2024-02-03
fix: ensure interactions are not left when exception occur inside with statement (#533)

### LINKED ISSUE #533: Interactions not cleared when error in test
### Have you read the Contributing Guidelines on issues?

- [ ] I have read the [Contributing Guidelines on issues](https://github.com/pact-foundation/pact-python/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of `pact-python`.
- [ ] I have read the console error message carefully (if applicable).

### Description

I have two tests with the same pact. When the first one raise an exception "verify" is not called and interactions are not cleared. When next test runs it sees other tests interactions and fails. When tests run in different order then there is no dependency issue.

### Reproducible demo

https://gist.github.com/dryobates/2a5b3028fc7f92dca23830b3ec0be225

### Steps to reproduce

pytest test_failing_pact_dependency.py

### Expected behavior

There should be no dependency between two tests.

### Actual behavior

Currently when one test raise an exception, the other test also fails but should not.

### Your environment

- Pact Python version used: master branch and 1.7.0
- OS: ArchLinux

### Self-service

- [X] I'd be willing to fix this bug myself.

dryobates:

## :memo: Summary

Interactions are cleared not only at "verify" but also at the time of error while leaving with statement.

## :fire: Motivation

It's a bug fix. Without it failure in one test can mess with the other test.

## :hammer: Test Plan

Run failing code submitted in the issue #533 

## :link: Related issues/PRs

closes #533

### DIFF
--- a/pact/pact.py
+++ b/pact/pact.py
@@ -383,6 +383,7 @@ class Pact(Broker):
         Calls the mock service to verify that all interactions occurred as
         expected, and has it write out the contracts to disk.
         """
+        self._interactions = []
         if (exc_type, exc_val, exc_tb) != (None, None, None):
             return
 
--- a/tests/test_pact.py
+++ b/tests/test_pact.py
@@ -562,6 +562,22 @@ class PactContextManagerTestCase(PactTestCase):
         self.mock_setup.assert_called_once_with(pact)
         self.assertFalse(self.mock_verify.called)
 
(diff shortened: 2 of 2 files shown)


JP-Ellis:
Great catch, and thanks for the detailed issue!

JP-Ellis (approved):

Issue #381, opened 2023-09-13
Groundwork API for Provider API using Rust backend


mefellows:
“Add support” means adding DSL methods that call the underlying Rust functionality.

“Add example” means create an example Python project that showcases using the functionality.

PR #575, opened 2024-02-29
Groundwork API for Provider

JP-Ellis:
## :memo: Summary

Groundwork to be able to verifier HTTP pacts.

## :fire: Motivation

Part of the work for V3

## :hammer: Test Plan

CI Workflow

## :link: Related issues/PRs

- #381

### DIFF
--- a/src/pact/v3/__init__.py
+++ b/src/pact/v3/__init__.py
@@ -23,9 +23,11 @@ be as follows:
 import warnings
 
 from pact.v3.pact import Pact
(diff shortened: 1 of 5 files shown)


JP-Ellis:
The FFI implies that the `verifier` is the core entity; though I am still on the fence whether I should call the class `Verifier` or `Provider`. Given that a verifier only ever has a single provider (though I guess a provider may have multiple verifiers), it might be more intuitive to rename the class.

PR #655, opened 2024-04-11
chore(ci): rework continue-on-error

JP-Ellis:
## :memo: Summary

There's no point testing on Python 3.13 as we can't even install hatch at this stage (let alone test Pact Python).

## :fire: Motivation

To get a workflow status that is properly reflective of the real status.

## :hammer: Test Plan

CI

## ~:link: Related issues/PRs~


JP-Ellis:
Closing as it isn't exactly required. The badge is marked as a success (even if the commit has a cross).

PR #761, opened 2024-08-13
adding matcher POC

valkolovos:
matcher POC

I just want to verify if this is headed in the right direction


examples/tests/v3/basic_flask_server.py
valkolovos, line 1: This is just to serve the expected response and would need to be heavily modified for a final product.

src/pact/v3/matchers/matchers.py
valkolovos, line 1: I implemented a generator option _only_ in the `regex` matcher again just as a POC.

valkolovos: The Matcher class will (I think) be useful when implementing more complex logic, just as being able to "OR" together different matchers. As far as I can tell, there's no way to do the "OR" logic directly in parameter JSON.

examples/tests/v3/test_matchers.py
```diff
@@ -0,0 +1,48 @@
+from pathlib import Path
+
+from examples.tests.v3.basic_flask_server import start_provider
+from pact.v3 import Pact, Verifier, matchers
```
JP-Ellis, line 4: I would want to avoid polluting the namespace, so I agree with your choice of importing a module and calling `matchers.like` and `matchers.regex` etc.

My nit here is that I would want to name it `match` so that the syntax is `match.regex(...)` and `match.like(...)`, etc. This way, it reads a bit more like English :)

src/pact/v3/interaction/_base.py
```diff
@@ -266,11 +267,16 @@ def with_body(
                 If `None`, then the function intelligently determines whether
                 the body should be added to the request or the response.
         """
+        if not isinstance(body, str):
+            body_str = json.dumps(body, cls=MatcherEncoder)
```
JP-Ellis, line 271: Nit: Avoid negative `isinstance` assertions.

If Python was strongly typed, this wouldn't matter; however, the type annotations are merely hint and at runtime, end users can in fact input whatever they want. So we need to handle cases of `with_body(MyClassWithouSerialization(...))` (which your `if` statement would catch).

JP-Ellis (commented):
Love the PoC!

I have a few minor nitpicks, but otherwise this looks like an amazing start!

Let me know how you would like me to support you in this going forward 🚀

examples/tests/v3/test_matchers.py
valkolovos: I feel like the module that defines the actual classes (Matcher, MatcherEncoder, etc) should be called "matchers" since they are objects. I'd be happy to create a module called match and move the functions that return those matchers into that module.

JP-Ellis: Just started going through your updated PR. I'm still thinking that we should expose the the functions behind a common `match.{func}` interface, as I do think it becomes most intuitive. It also serves as a nice way to distinguish `match.{func}` from similar `generate.{func}`.

I also agree that having `matchers` as the module name for the Matcher classes makes sense. Having said that, I don't think people would often (or ever?) need to interact with the class directly. So I'm liking what you've done whereby the `matchers` are defined in a submodule, and functions re-exported at a higher level.

JP-Ellis:
I've started working on the PR. Pushed a few changes (which are so far relatively minor).

I do anticipate making some larger changes at a later date, but I just want to get a feel for it first :)

src/pact/v3/match/types.pyi
valkolovos, line 1: TIL about stub files :)

src/pact/v3/generate/generator.py
```diff
@@ -0,0 +1,161 @@
...
+    additional configuration elements.
+    """
+
+    def __init__(
+        self,
+        type: GeneratorType,  # noqa: A002
+        /,
+        extra_fields: Mapping[str, Any] | None = None,
+        integration_fields: Mapping[str, Any] | None = None,
+        generator_json_fields: Mapping[str, Any] | None = None,
```
valkolovos, line 83: Is it possible to provide examples of how these two might be used?

JP-Ellis: Yeah, good call.

The reasoning initially is that the way the generators/matchers are serialised into the 'integration JSON' format is slightly different to the way they are serialised when passed directly to the `matching_rules` or `generators` commands. E.g., in one, you have to use `"pact:matcher:type": "..."` whereas the other you don't.

As far as I know, all the other fields are the same, but if there were others that had to be differentiated, then this is there.

Thinking back on it though, I'm concerned that this might be a case of premature generalisation.

JP-Ellis:
Even though this is failing on Python 3.8, I will still merge this as Python 3.8 is being dropped imminently (#796)

PR #773, opened 2024-08-27
chore: added v3 http interaction examples

amit828as:

-   [x ] I have read the [Contributing Guidelines on pull requests](https://github.com/pact-foundation/pact-python/blob/master/CONTRIBUTING.md#pull-requests).
-   [x ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
-   [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## :memo: Summary

This PR has example usage of Pact-python version 3 HTTP interaction


examples/tests/v3/test_v3_http_consumer.py
```diff
@@ -0,0 +1,94 @@
+"""
+This module provides test cases and example usage for the HTTP consumer in version 3.
```
JP-Ellis, line 2: This will need to be expanded significantly to ensure that the examples serve their pedagogical purpose.

examples/tests/v3/test_v3_http_consumer.py
```diff
@@ -0,0 +1,94 @@
...
+    ```
+    Returns a generator that yields the pact object and writes the pact file
+    """
+    pact_dir = Path(Path(__file__).parent.parent.parent / "pacts")
+    pact = Pact("v3_http_consumer", "v3_http_provider")
+    yield pact.with_specification("V3")
+    pact.write_file(pact_dir)
+
+
+# Example 1: GET request for existing user
```
JP-Ellis, line 42: I would get rid of this comment, and rename the test function to a more descriptive (and self-explanatory) `test_get_existing_user`.

examples/tests/v3/test_v3_http_consumer.py
```diff
@@ -0,0 +1,94 @@
...
+    Test the GET request for retrieving user information.
+
+    This test defines the expected interaction for a GET request to retrieve
+    user information. It sets up the expected request and response from the
+    provider and verifies that the response status code is 200.
+    """
+    # When setting up the expected response, the consumer should only define
+    # what it needs from the provider (as opposed to the full schema). Should
+    # the provider later decide to add or remove fields, Pact's consumer-driven
+    # approach will ensure that interaction is still valid.
```
JP-Ellis, line 54: I would move this comment into the docstring (and if it wasn't in the docstring elsewhere, I would move it there too).

The docstring gets picked up by `mkdocs` and go into the docs website, unlike the in-code comments. So putting more in the docstrings ensure there is more visibility.

examples/tests/v3/test_v3_http_consumer.py
```diff
@@ -0,0 +1,94 @@
...
+        200  # In this example, we only care about the response code
+    )
+    (
+        pact.upon_receiving(
+            "a request for user information"
+        )  # Defining the interaction
+        .given("user exists")  # Defining the provider state
+        .with_request(method="GET", path="/users/1")
+        .will_respond_with(200)
+    )
```
JP-Ellis, line 66: Could we test the response body as well?

I recall the existing examples showcasing how consumer tests only verify the _minimal_ response required. I.e., the provider might be responding with superfluous data which the consumer does not care about, and Pact should be testing only the what is strictly necessary for the consumer to work, as opposed to the entire provider payload.

examples/tests/v3/test_v3_http_consumer.py
```diff
@@ -0,0 +1,94 @@
...
+
+    with pact.serve() as srv:
+        response = requests.get(f"{srv.url}/users/1", timeout=5)
+
+        assert response.status_code == expected_response_code
+
+
+# Example 2: GET request for non-existing user
```
JP-Ellis, line 74: As above, the comment seems a bit superfluous when the function name is self-explanatory.

examples/tests/v3/test_v3_http_consumer.py
```diff
@@ -0,0 +1,94 @@
...
+    This test defines the expected interaction for a GET request to retrieve
+    user information. It sets up the expected request and response from the
+    provider and verifies that the response status code is 400.
+    """
+    expected_response_code = 404
+    (
+        pact.upon_receiving("a request for user information")
+        .given("user doesn't exists")
+        .with_request(method="GET", path="/users/2")
+        .will_respond_with(404)
```
JP-Ellis, line 88: You don't include a body in the response.

This actually is probably fine, especially if the client implementation short-circuits the parsing of the body if it first detects an error status code.

However, the docstring should be updated to make it very clear as to _why_ we aren't testing for a body, and how that actually means that _any_ data in the body would be deemed as valid.

examples/tests/v3/test_v3_http_provider.py
```diff
@@ -0,0 +1,118 @@
...
+with the expected responses.
+
+The provider will be expected to be in a given state in order to respond to
+certain requests. For example, when fetching a user's information, the provider
+will need to have a user with the given ID in the database. In order to avoid
+side effects, the provider's database calls are mocked out using functionalities
+from `unittest.mock`.
+
+In order to set the provider into the correct state, this test module defines an
+additional endpoint on the provider, in this case `/_pact/provider_states`.
```
JP-Ellis, line 16: This has been renamed to `/_pact/callback` below.

examples/tests/v3/test_v3_http_provider.py
```diff
@@ -0,0 +1,118 @@
...
+    For Pact to be able to correctly get the provider into the correct state,
+    this function is used to define an additional endpoint on the provider. This
+    endpoint is called by Pact before each test to ensure that the provider is
+    in the correct state.
+    """
+    mapping = {
+        "user doesn't exists": mock_user_doesnt_exist,
+        "user exists": mock_user_exists,
+    }
+    return {"result": mapping[state]()}
```
JP-Ellis, line 60: I actually think we should handle the teardown in this situation.

As part of the teardown, I would want to check that the mocked database had the expected calls made to it.

For example, if we are expecting the request to perform a `FAKE_DB.get(1)`, then as part of the teardown, we should probably verify that the `get` method was indeed called, with the expected argument.

This would require the addition of two new functions: `verify_user_doesnt_exist_mock` and `verify_user_exists_mock`, and we would need to create a separate mapping based on whether the setup or teardown callback action is being sent.

examples/tests/v3/test_v3_http_provider.py
```diff
@@ -0,0 +1,118 @@
...
+    cannot be pickled.
+    """
+    host = PROVIDER_URL.host if PROVIDER_URL.host else "localhost"
+    port = PROVIDER_URL.port if PROVIDER_URL.port else 8000
+    uvicorn.run(app, host=host, port=port)
+
+
+def test_provider() -> None:
+    """
+    Test the provider to ensure compliance.
```
JP-Ellis, line 78: Since this is the _only_ test function, the docstring for this function should be expanded significantly to explain exactly what is going on 'under the hood'.

JP-Ellis (changes requested):
Looks good!

There's a few minor things I think should be changed to make sure that the examples serve as a good learning tool for people who are otherwise unfamiliar with Pact.

There is also a bit of a change in the provider testing which I think is needed. While Pact performs the necessary verifications on the server's communications, it would be good to show how the test can be used to verify that side-effects have (or have not) taken place within the provider's internal logic.

Let me know what you think, and if you have any questions, let me know!

JP-Ellis (approved):
Looks good 🚀 

Just cleaned up a few little things, mostly around the docs and typing 👍

JP-Ellis:
There's an upstream with CFFI on Windows:

- [ ] python-cffi/cffi#117

Will do a force merge later today as all the other tests are fine.

PR #796, opened 2024-09-20
feat!: add Python 3.13, drop 3.8

JP-Ellis:
## :memo: Summary

With the release of Python 3.13 and Python 3.8 no longer being maintained, we update the supported Python versions of Pact Python to match what is currently maintained.

Python 3.8 support is dropped

## :fire: Motivation

To keep align with the supported versions of Python

## :hammer: Test Plan

CI

## :link: Related issues/PRs

None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  STABLE_PYTHON_VERSION: "3.12"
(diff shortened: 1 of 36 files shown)


JP-Ellis:
This should be ready to merge as soon as the [actions/setup-python](https://github.com/actions/setup-python) action supports 3.13 (which is actually done through the data stored in the [`versions-manifest.json`](https://github.com/actions/python-versions/blob/main/versions-manifest.json))

Issue #644, opened 2024-04-02
Add structured `Mismatch` class


JP-Ellis:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/pact-foundation/pact-python/blob/master/CONTRIBUTING.md#issues).

### Description

Replace the generic `dict[str, Any]` returned by a mismatch with a more informative class:

https://github.com/pact-foundation/pact-python/blob/5e52872b46fd6f959e958c130ebbe08ef8e11bbc/src/pact/v3/pact.py#L407-L412

### Motivation

A new class would facilitate the programmatic handling of mismatches, and avoid the use of `Any`.

### Self-service

- [ ] I'd be willing to contribute this feature to Pact Python myself.

PR #815, opened 2024-10-08
feat: improve mismatch error

JP-Ellis:
## :memo: Summary

Adds a `Mismatch` class and a number of subclasses.

This is done to facilitate the error handling, and error displays.

- The type of the mismatches in the `MismatchesError` is now different. It was a `dict[str, Any]` and is now a `Mismatch` sub-class.

## :fire: Motivation

Improve error messages, as the old error messages were clearly not appropriate.

## :hammer: Test Plan

At time of writing, yet to be done.

## :link: Related issues/PRs

- Resolves: #644

### DIFF
--- a/src/pact/v3/error.py
+++ b/src/pact/v3/error.py
@@ -1,0 +1,1117 @@
+"""
+Error classes for Pact.
+"""
+
+from __future__ import annotations
+
+import copy
+import logging
+from abc import ABC, abstractmethod
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+
+class PactError(Exception, ABC):
+    """
+    Base class for exceptions raised by the Pact module.
+    """
+
+
+class InteractionVerificationError(PactError):
+    """
+    Exception raised due during the verification of an interaction.
+
+    This error is raised when an error occurs during the manual verification of an
+    interaction. This is typically raised when the consumer fails to handle the
+    interaction correctly thereby generating its own exception. The cause of the
+    error is stored in the `error` attribute.
+    """
+
+    def __init__(self, description: str, error: Exception) -> None:
+        """
+        Initialise a new InteractionVerificationError.
+
+        Args:
+            description:
+                Description of the interaction that failed verification.
+
+            error: Error that occurred during the verification of the
+                interaction.
+        """
+        super().__init__(f"Error verifying interaction '{description}': {error}")
+        self._description = description
+        self._error = error
+
+    @property
+    def description(self) -> str:
+        """
+        Description of the interaction that failed verification.
+        """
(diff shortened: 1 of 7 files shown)


JP-Ellis:
@individual-it This hopefully goes some way to address the visibility of the mismatches when a Pact test fails.

I'm still implementing the tests for this, but the actual implementation should mostly be there (pending confirmation from the tests I'm writing)

src/pact/v3/error.py
```diff
@@ -0,0 +1,831 @@
...
+        Create a new Mismatch from a dictionary.
+
+        Args:
+            data:
+                Data for the mismatch.
+
+        Returns:
+            A new Mismatch object.
+        """
+        if mismatch_type := data.pop("type"):
```
individual-it, line 123: when running into this error with consumer tests, `data` is a list not a dict
e.g. 
```
[{'method': 'GET', 'path': '/users/123', 'request': {'body': '', 'headers': {'accept': '*/*', 'accept-encoding': 'gzip, deflate, zstd', 'connection': 'keep-alive', 'host': 'localhost:38499', 'user-agent': 'python-requests/2.32.3'}, 'method': 'GET', 'path': '/users/123'}, 'type': 'request-not-found'}, {'method': 'GET', 'path': '/user/123', 'request': {'method': 'GET', 'path': '/user/123'}, 'type': 'missing-request'}]
```

so this ends up in a `TypeError: 'str' object cannot be interpreted as an integer`

src/pact/v3/error.py
```diff
@@ -0,0 +1,831 @@
...
+            if mismatch_type == "QueryMismatch":
+                return QueryMismatch(**data)
+            if mismatch_type == "HeaderMismatch":
+                return HeaderMismatch(**data)
+            if mismatch_type == "BodyTypeMismatch":
+                return BodyTypeMismatch(**data)
+            if mismatch_type == "BodyMismatch":
+                return BodyMismatch(**data)
+            if mismatch_type == "MetadataMismatch":
+                return MetadataMismatch(**data)
```
individual-it, line 139: I could not provoke any of those types.
All types I could provoke so far (in consumer tests) are Kebab Case
`request-not-found` or `missing-request`
so they will always end up as `GenericMismatch`. Is that intended?

individual-it (commented):
Thank you!
I has no time yet to try that with provider side tests, but with consumer side tests I found some issues in the `from_dict` function

JP-Ellis:
Yeah I'm still working on the breaking changes introduced for tests that inspect errors. It's also catching a number of other 'mismatches' such as the request not found you mentioned.

PR #890, opened 2024-12-05
Allow functional arguments for provider callbacks and message producers

JP-Ellis:
## :memo: Summary

Allows to functions to be passed in to handle provider states and to produce messages.

Still outstanding at the time of writing this PR is fully updated documentation.

As part of these changes,

This is all within the `pact.v3` module which is not yet stable. The breaking changes are as follows:

- `Verifier.set_info` has been removed. Its functionality has been split across two functions:
  - `Verifier.__init__` which takes two arguments now: the provider name and the host against which to run the provider tests.
  - `Verifier.add_transport` which behaves as before but now _needs_ to be called at least once before executing `.verify()`

- `Verifier.set_state` has been replaced by `Verifier.state_handler`. This more correctly reflects the intent of the method.

## :fire: Motivation

Previously, if one wanted to have a function be called to set the provider in a specific state, the end-user implementing the test suite would need to write a minimal HTTP server to effectively wrap this function.

This capability is still present, but it is now possible to pass in a function directly. Pact Python internally handles the HTTP server wrapper so the end user need not worry about that.

Similarly, the message producer required a HTTP wrapper to trigger the message generation. This has been made automatic with the end-user only requiring to define a function to handle that.

## :hammer: Test Plan

All existing tests have been adapted to make use of this feature.

## :link: Related issues/PRs

- Resolves: #749


JP-Ellis:
@valkolovos Here's the PR I have been talking about. Keen to have you give it a go and tell me what you think 🚀

examples/tests/v3/test_01_fastapi_provider.py
```diff
@@ -18,99 +18,79 @@
 side effects, the provider's database calls are mocked out using functionalities
 from `unittest.mock`.
 
-In order to set the provider into the correct state, this test module defines an
-additional endpoint on the provider, in this case `/_pact/callback`. Calls to
-this endpoint mock the relevant database calls to set the provider into the
-correct state.
+Note that Pact requires tat the provider be running on an accessible URL. This
```
valkolovos, line 21: nit - typo... should be "requires that"

examples/tests/v3/test_01_fastapi_provider.py
```diff
@@ -154,18 +134,68 @@ def test_provider() -> None:
...
+
+def provider_state_handler(
+    state: str,
+    action: str,
+    _parameters: dict[str, Any] | None,
+) -> None:
+    """
+    Handler for the provider state callback.
+
+    For Pact to be able to correctly tests compliance with the contract, the
```
valkolovos, line 154: another typo nit - "to correctly test compliance"

PR #914, opened 2024-12-30
docs(blog): add functional arguments post

JP-Ellis:
## :memo: Summary

Version 2.3.0 of Pact Python adds support for functional arguments in a few Verifier methods. This blog posts documents the changes and how they can be used. It also documents the breaking changes that were introduced.

## :fire: Motivation

To help those users using the `pact.v3` module already, despite it not being stable.

## :link: Related issues/PRs

- #890


vwong (approved):
I haven't tried the before/after changes on real-world code, but it all seems fine to me.

Issue #1084, opened 2025-06-20
Add example using 'protobuf' plugin


JP-Ellis:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/pact-foundation/pact-python/blob/main/CONTRIBUTING.md#issues).

### Description

Adding a plugin which makes use of the protobuf plugin.

### Motivation

To help with the adoption of plugins.

### Have you tried building it?

WIP

### Self-service

- [x] I'd be willing to contribute this feature to Pact Python myself.

PR #1089, opened 2025-06-25
chore(examples): add protobuf example

JP-Ellis:
## :memo: Summary

Add a new example showcasing how to use the Protobuf plugin to validate Protobuf over HTTP content.

This first example is crude and does not yet do any content matching. This will be done in a follow up PR.

## :fire: Motivation

Plugins are not very easy to use, and the `v3` API is still very new. This should hopefully help with adoption of V3 and help end-users of Pact Python adopt the tool.

## :hammer: Test Plan

All examples are tested as part of CI.

## :link: Related issues/PRs

- [ ] #1084


examples/v3/plugins/protobuf/test_consumer.py
```diff
@@ -0,0 +1,140 @@
...
+    The provider state ensures that a person with ID 1 exists in the system,
+    corresponding to Alice from our sample address book.
+    """
+    sample_address_book = address_book()
+    alice = sample_address_book.people[0]
+    expected_protobuf_data = alice.SerializeToString()
+
+    (
+        pact.upon_receiving("a request to get person by ID")
+        .given("person with ID 1 exists")
```
YOU54F, line 81: would it be better to use parameterised provider states here?

examples/v3/plugins/protobuf/test_provider.py
```diff
@@ -0,0 +1,247 @@
...
+
+        action:
+            Either `"setup"` or `"teardown"`.
+
+        parameters:
+            Additional parameters (not used in this example).
+    """
+    if action == "setup":
+        {
+            "person with ID 1 exists": setup_person_exists,
```
YOU54F, line 201: can we have an example using provider state params here? as per previous comment

YOU54F (approved):
looks good, comment about adding parameterised provider states, but other than that 👌🏾 from me. code docs are useful for context

Issue #1107, opened 2025-07-15
V3: wip pacts not found because of missing property providerVersionBranch


patrickbouwmeester:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/pact-foundation/pact-python/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of `pact-python`.
- [x] I have read the console error message carefully (if applicable).

### Description

We have been working for a while with pact python V3 and now we got a small issue with V3 when we published a contract for the first time to pact broker and we want to verify this contract for the first time (this contract is still in a MR and not deployed to our environments). The consumer version selectors we use are: "mainBranch : True", "deployedOrReleased: True", include_pending and include_wip_since("2025-07-11"). The result of the test is that no pacts were found with these selectors while we expect that the contract must be found because of the selector wip since. When we  use pact python V2 and use the same selectors then the pact contract is found in pact broker. 

Here you see the code we use to test the contract with pact-broker

```python
verifier = (
    Verifier(PROVIDER)
    .add_transport(url=server)
    .state_handler(provider_state_handler, teardown=False)
    .broker_source(url=BROKER_URL, selector=True)
    .consumer_versions('{"mainBranch": true}', '{"deployedOrReleased": true}')
    .include_pending()
    .include_wip_since("2025-07-11")
    .build
)

if os.getenv("CI"):
    verifier.set_publish_options(
        version=os.getenv("VERSION")
        branch=os.getenv("CI_COMMIT_REF_NAME")
    )

verifier.verify()
```

After some debugging we see that there is a small difference in the request that's send for the POST call for url "…/for-verification". 

The json created by v3:
```json
{
   "includePendingStatus": true,
   "includeWipPactsSince": "2025-07-11",
   "consumerVersionSelectors": [
       {
           "mainBranch": true
       },
       {
           "deployedOrReleased": true
       }
   ]
}
```

The json created by v2:
```json
{
    "includePendingStatus": true,
    "consumerVersionSelectors": [
        {
            "mainBranch": true
        },
        {
            "deployedOrReleased": true
        }
    ],
    "providerVersionBranch": "INT-7657",
    "includeWipPactsSince": "2025-07-11"
}
```

[...]

PR #1162, opened 2025-08-08
chore: set verifier branch same as publisher branch

kevinrvaz:

-   [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## :memo: Summary

In most cases, a provider will want to verify Pacts against a branch, and publish against the same branch. Despite this, the branch in the broker selector and the branch in the publish options were entirely unrelated.

This PR now ensures they are linked so that they can be set through either option, and re-used automatically in the other.

If a branch is set through either `set_publish_options` or `provider_branch`, the value will be saved and used as a default for both in subsequent calls. 

This is unlikely to impact end-users, and a `None` value can be set explicitly.

## :fire: Motivation

Ensures consistency between `set_publish_options` and the broker selectors.

## :link: Related issues/PRs

Fixes: #1107

### DIFF
--- a/src/pact/verifier.py
+++ b/src/pact/verifier.py
@@ -166,6 +166,7 @@ class Verifier:
         self._name = name
         self._host = host or "localhost"
         self._handle = pact_ffi.verifier_new_for_application()
+        self._branch: str | None = None
 
         # In order to provide a fluent interface, we remember some options which
         # are set using the same FFI method. In particular, we remember
@@ -864,13 +865,19 @@ class Verifier:
 
             branch:
                 Name of the branch used for verification.
+
+                The first time a branch is set here or through
+                [`provider_branch`][pact.verifier.BrokerSelectorBuilder.provider_branch],
+                the value will be saved and use as a default for both.
         """
+        if not self._branch and branch:
+            self._branch = branch
         pact_ffi.verifier_set_publish_options(
             self._handle,
             version,
             url,
             tags or [],
-            branch,
+            branch or self._branch,
         )
         return self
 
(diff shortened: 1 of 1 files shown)


src/pact/verifier.py
```diff
@@ -198,6 +198,12 @@ def __repr__(self) -> str:
         """
         return f"<Verifier: {self._name}, handle={self._handle}>"
 
+    def branch(self, branch: str | None) -> None:
+        """
+        Set branch string to be used with provider.
+        """
+        self._branch = branch
+
```
JP-Ellis: I wouldn't create a new public method just for this, since it doesn't really do much on its own that we couldn't do 'behind the scenes' (and in fact, do behind the scenes anyway).

I don't see a situation where someone would call `Verifier.branch` and not call the other options.

src/pact/verifier.py
```diff
@@ -814,6 +820,7 @@ def set_publish_options(
             tags or [],
             branch,
         )
+        self.branch(branch)
```
JP-Ellis: We should conditionally update this if it isn't set already:

```suggestion
        if not self._branch and branch:
            self._branch = branch
```

and then the branch argument in `set_publish_options` should be changed to `branch or self._branch`. This ensures that the explicit value is always used before falling back.

Lastly, the documentation for the `branch` argument should make it clear how the default can be set (and link to it). E.g. something along the lines of:

```
If the branch is unspecified, the value will be populated from the [`provider_branch`][...] argument.
```

src/pact/verifier.py
```diff
@@ -1298,6 +1305,7 @@ def provider_branch(self, branch: str) -> Self:
         Set the provider branch.
         """
         self._provider_branch = branch
+        self._verifier.branch(branch)
```
JP-Ellis: As above, this should override the save value only if it isn't set already:

```
if not self._verifier._branch:
    self._verifier._branch = branch
```

And the docstring should be updated to make it clear that if this value os omitted, it is taken from the `publish_options` argument.

src/pact/verifier.py
kevinrvaz: removed

src/pact/verifier.py
kevinrvaz: updated

src/pact/verifier.py
kevinrvaz: updated

JP-Ellis (approved):

PR #1253, opened 2025-09-29
docs: update for v3 and add migration guide

JP-Ellis:
## :memo: Summary

Overhaul of the documentation in preparation of the V3 release.

In particular:

- Add a migration guide
- Update the consumer and provider docs with the new V3 interface
- Update the matcher and generator module docs

The abstract `Matcher` and `Generator` classes have been renaned to `AbstractMatcher` and `AbstractGenerator`. This is not going to be a breaking change unless you have created your own matcher or generator.

## :fire: Motivation

To help with the adoption of the soon-to-be-released version 3, for both new and existing Pact Python users.

## :link: Related issues/PRs

- Resolves: #1222

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -103,7 +103,9 @@ The v3 interface favours method chaining and provides more granular control over
     .with_body({'id': 123, 'name': 'Alice'}, content_type='application/json'))
 ```
 
-1.  The new provider states can be streamlined by parameterizing them directly in the `given()` method. So instead of defining multiple variations of a `"user exists"` state, you can define it once and pass different parameters as needed. These can be passed as keyword arguments to `given()`, or as a dictionary in the second positional argument.
+1.  In v2, there was limited support for parameterizing provider states, and each state variation often required a separate definition. For example, `given("user Alice exists with id 123")` and `given("user Bob exists with id 456")` would be two distinct states, which would then need to be handled separately in the provider state setup.
+
+    The new interface can now define a common descriptor that can be reused with different parameters: `.given("user exists", id=123, name='Alice')` and `.given("user exists", id=456, name='Bob')`. This approach reduces redundancy and makes it easier to manage provider states.
 
 Some methods are shared across request and response definitions, such as `with_header()` and `with_body()`. Pact Python automatically applies them to the correct part of the interaction based on whether they are called before or after `will_respond_with()`. Alternatively, these methods accept an optional `part` argument to explicitly specify whether they apply to the request or response.
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,0 +1,425 @@
+# Migration Guide
+
+> [!INFO]
+>
+> This document is best viewed on the [Pact Python docs site](https://pact-foundation.github.io/pact-python/MIGRATION/).
+
+This document outlines the key changes and migration steps for users transitioning between major Pact Python versions. It focuses on breaking changes, new features, and best practices to ensure a smooth upgrade.
+
+## Migrating from `2.x` to `3.x`
+
+### Key Changes
+
+-   Replaced the entire Pact CLI-based implementation with a brand new version leveraging the [core Pact FFI](https://github.com/pact-foundation/pact-reference) written in Rust.
+    -   This will help ensure feature parity between different language implementations and improve performance and reliability. This also brings compatibility with the latest Pact Specification (v4).
+-   The bundled CLI is now a separate package: [`pact-python-cli`](https://pypi.org/project/pact-python-cli/).
+-   The programmatic API has been completely overhauled to be more Pythonic and easier to use. Legacy process calls and return code checks have been removed in favour of proper exception handling.
+-   The old programmatic interface is available in the `pact.v2` module for backwards compatibility. This is deprecated, will not receive new features, and will be removed in a future release.
+
+### Using the `pact.v2` Compatibility Module
+
+For teams with larger codebases that need time to fully migrate to the new v3 API, a backwards compatibility module is provided at `pact.v2`. This module contains the same API as Pact Python v2.x and serves as an interim measure to assist gradual migration.
+
+To use the v2 compatibility module, you must install pact-python with the `v2` feature enabled:
+
+```bash
+pip install pact-python[v2]
+```
+
+All existing `pact.*` imports need to be updated to use `pact.v2.*` instead. Here are some common examples:
+
+```python
+# Old v2.x imports
+from pact import Consumer, Provider
+from pact.matchers import Like, EachLike
+from pact.verifier import Verifier
+
+# New v3.x imports using the v2 compatibility module
+from pact.v2 import Consumer, Provider
+from pact.v2.matchers import Like, EachLike
+from pact.v2.verifier import Verifier
+```
+
+Please note that this compatibility module is intended as a temporary solution. A full migration to the new v3 API is strongly recommended as soon as feasible. The compatibility module will only receive critical bug fixes and no new features.
+
+Pact, by default, updates existing Pact contracts in place, so migrating consumer tests incrementally should be feasible. However, newer features (e.g., message pacts) will likely require a full migration, and mixed usage of v2 and v3 APIs is not officially supported.
+
+### Consumer Changes
+
+The v3 API introduces significant changes to how consumer tests are structured and written. The main changes simplify the API, making it more Pythonic and flexible.
+
+#### Defining a Pact
+
+The `Consumer` and `Provider` classes have been removed. Instead, a single `Pact` class is used to define the consumer-provider relationship:
+
+```python title="v2"
+from pact.v2 import Consumer, Provider
+
+consumer = Consumer('my-web-front-end')
+provider = Provider('my-backend-service')
+
+pact = consumer.has_pact_with(provider, pact_dir='/path/to/pacts')
+```
+
+```python title="v3"
+from pact import Pact
+
+pact = Pact('my-web-front-end', 'my-backend-service')
+```
+
+#### Defining Interactions
+
+The v3 interface favours method chaining and provides more granular control over request and response definitions.
+
+```python title="v2"
+(
+    pact
+    .given('user exists')
+    .upon_receiving('a request for user data')
+    .with_request(
+        'GET',
+        '/users/123',
+        headers={'Accept': 'application/json'},
+        query={'include': 'profile'}
+    )
+    .will_respond_with(
+        200,
+        headers={'Content-Type': 'application/json'},
+        body={'id': 123, 'name': 'Alice'}
+    )
+)
+```
+
+```python title="v3"
+(
+    pact
+    .upon_receiving('a request for user data')
+    .given('user exists', id=123, name='Alice')  # (1)
+    .with_request('GET', '/users/123')
+    .with_header('Accept', 'application/json')
+    .with_query_parameter('include', 'profile')
+    .will_respond_with(200)
+    .with_header('Content-Type', 'application/json')
+    .with_body({'id': 123, 'name': 'Alice'}, content_type='application/json'))
+```
+
+1.  The new provider states can be streamlined by parameterizing them directly in the `given()` method. So instead of defining multiple variations of a `"user exists"` state, you can define it once and pass different parameters as needed. These can be passed as keyword arguments to `given()`, or as a dictionary in the second positional argument.
+
+Some methods are shared across request and response definitions, such as `with_header()` and `with_body()`. Pact Python automatically applies them to the correct part of the interaction based on whether they are called before or after `will_respond_with()`. Alternatively, these methods accept an optional `part` argument to explicitly specify whether they apply to the request or response.
+
+#### Running Tests
+
+Pact Python v2 had two different ways to run consumer tests, both of which spawned a separate mock service process. The new v3 API provides a single, consistent way to run tests using the `serve()` method.
+
+```python title="v2 - with context manager"
+pact = Consumer("my-consumer").has_pact_with(
+    Provider("my-provider"),
+    host_name="localhost",
+    port=1234,
+)
+
+# Context manager automatically calls setup() and verify()
+with pact:
+    response = requests.get(pact.uri + '/users/123')
+# Pact file written automatically on exit
+```
+
+```python title="v2 - with manual service management"
+pact = Consumer("my-consumer").has_pact_with(
+    Provider("my-provider"),
+    host_name="localhost",
+    port=1234,
+)
+
+# Manually start the mock service
+pact.start_service()
+pact.setup()  # Configure interactions
+
+# Make requests
+response = requests.get(pact.uri + '/users/123')
+# Assertions...
+
+# Verify and stop
+pact.verify()  # Writes pact file
+pact.stop_service()
+```
+
+The new API entirely replaces both of these approaches with a single, consistent method:
+
+```python title="v3"
+pact = Pact("my-consumer", "my-provider")
+
+with pact.serve() as srv:
+    response = requests.get(f"{srv.url}/users/123")
+```
+
+The server host and port can be specified in `serve()` if needed, but by default, the server binds to `localhost` on a random available port. More details can be found in the [API reference][pact.pact.Pact.serve].
+
+##### Writing Pact Files
+
+Since the old v2 API executed a sub-process for the mock service, the Pact file was automatically written when the context manager exited or when `pact.verify()` was called. The new v3 API runs the mock service in-process, so the Pact file must be written explicitly using the `write_file()` method:
+
+```python title="v2"
+with pact:
+    # tests...
+# Pact file written automatically
+```
+
+```python title="v3"
+pact = Pact('consumer', 'provider')
+# Define interactions and run tests...
+pact.write_file('/path/to/pacts')
+```
+
+#### Matchers
+
+Support for matchers has been greatly expanded and improved in the v3 API. The older v2 classes defined a limited set of matchers, while the new API provides a more comprehensive and flexible approach.
+
+```python title="v2"
+from pact.v2.matchers import Like, EachLike, Regex, Term
+
+# Usage:
+Like({'id': 123})
+EachLike({'item': 'value'})
+Regex('hello world', r'^hello')
+```
+
+```python title="v3"
+from pact import match
+
+# Usage:
+match.like({'id': 123})
+match.each_like({'item': 'value'})
+match.regex('hello world', r'^hello')
+```
+
+For a full list of available matchers and their usage, refer to the [API documentation][pact.match].
+
+### Provider Changes
+
+The provider verification API has been completely redesigned in v3 to provide a more intuitive and flexible interface. The old `Provider` and `Verifier` classes have been replaced by a single `Verifier` class with a fluent interface.
+
+#### Creating a Verifier
+
+```python title="v2"
+from pact.v2 import Provider, Verifier
+
+# Create separate Provider and Verifier instances
+provider = Provider('my-provider')
+verifier = Verifier(provider, 'http://localhost:8080')
+```
+
+```python title="v3"
+from pact import Verifier
+
+# Single Verifier instance with provider name
+verifier = Verifier('my-provider')
+```
+
+The protocol specification is now done through the `add_transport` method, which allows for more flexible configuration and supports multiple transports if needed.
+
+```python title="v2"
+verifier = Verifier(provider, 'http://localhost:8080')
+```
+
+```python title="v3"
+verifier = (
+    Verifier('my-provider')
+    .add_transport(url='http://localhost:8080')
+    # Or more granular control:
+    .add_transport(
+        protocol='http',
+        port=8080,
+        path='/api/v1',
+        scheme='https'
+    )
+)
+```
+
+#### Adding Pact Sources
+
+Support for both local files and Pact Brokers is retained in v3, with the `verify_pacts` and `verify_with_broker` methods replaced by a more flexible source configuration. This allows multiple sources to be combined, and selectors to be applied.
+
+<!-- markdownlint-disable code-block-style -->
+
+=== "Local Files"
+
+    ```python title="v2"
+    success, logs = verifier.verify_pacts(
+        './pacts/consumer1-provider.json',
+        './pacts/consumer2-provider.json'
+    )
+    ```
+
+    ```python title="v3"
+    verifier = (
+        Verifier('my-provider')
+        # It can discover all Pact files in a directory
+        .add_source('./pacts/')
+        # Or read individual files
+        .add_source('./pacts/specific-consumer.json')
+    )
+    ```
+
+=== "Pact Broker"
+
+    ```python title="v2"
+    success, logs = verifier.verify_with_broker(
+        broker_url='https://pact-broker.example.com',
+        broker_username='username',
+        broker_password='password'
+    )
+    ```
+
+    ```python title="v3"
+    verifier = (
+        Verifier('my-provider')
+        .broker_source(
+            'https://pact-broker.example.com',
+            username='username',
+            password='password'
+        )
+    )
+
+    # Or with selectors for more control
+    broker_builder = (
+        verifier
+        .broker_source(
+            'https://pact-broker.example.com',
+            selector=True
+        )
+        .include_pending()
+        .provider_branch('main')
+        .consumer_tags('main', 'develop')
+        .build()
+    )
+    ```
+
+    The `selector=True` argument returns a [`BrokerSelectorBuilder`][pact.verifier.BrokerSelectorBuilder] instance, which provides methods to configure which pacts to fetch. The `build()` call finalizes the configuration and returns the `Verifier` instance which can then be further configured.
+
+<!-- markdownlint-enable code-block-style -->
+
+#### Provider State Handling
+
+The old v2 API required the provider to expose an HTTP endpoint dedicated to handling provider states. This is still supported in v3, but there are now more flexible options, allowing Python functions (or mappings of state names to functions) to be used instead.
+
+<!-- markdownlint-disable code-block-style -->
+
+=== "URL-based State Handling"
+
+    ```python title="v2"
+    success, logs = verifier.verify_pacts(
+        './pacts/consumer-provider.json',
+        provider_states_setup_url='http://localhost:8080/_pact/provider_states'
+    )
+    ```
+
+    ```python title="v3"
+    # Option 1: URL-based (similar to v2)
+    verifier = (
+        Verifier('my-provider')
+        .add_transport(url='http://localhost:8080')
+        .state_handler(
+            'http://localhost:8080/_pact/provider_states',
+            body=True  # (1)
+        )
+        .add_source('./pacts/')
+    )
+    ```
+
+    1.  The `body` argument specifies whether to use a `POST` request and pass information in the body, or to use a `GET` request and pass information through HTTP headers. For more details, see the [`state_handler` API documentation][pact.verifier.Verifier.state_handler].
+
+=== "Functional State Handling"
+
+    ```python title="v2"
+    # Not supported
+    ```
+
+    ```python title="v3 - Function"
+        def handler(name, params=None):
+            if name == 'user exists':
+                # Set up user in database/mock
+                create_user(params.get('id', 123))
+            elif name == 'no users exist':
+                # Clear users
+                clear_users()
+
+        verifier = (
+            Verifier('my-provider')
+            .add_transport(url='http://localhost:8080')
+            .state_handler(handler)
+            .add_source('./pacts/')
+        )
+    ```
+
+    ```python title="v3 - Mapping"
+    state_handlers = {
+        'user exists': lambda name, params: create_user(params.get('id', 123)),
+        'no users exist': lambda name, params: clear_users(),
+    }
+
+    verifier = (
+        Verifier('my-provider')
+        .add_transport(url='http://localhost:8080')
+        .state_handler(state_handlers)
+        .add_source('./pacts/')
+    )
+    ```
+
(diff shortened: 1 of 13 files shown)


MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+    .with_header('Accept', 'application/json')
+    .with_query_parameter('include', 'profile')
+    .will_respond_with(200)
+    .with_header('Content-Type', 'application/json')
+    .with_body({'id': 123, 'name': 'Alice'}, content_type='application/json'))
+```
+
+Some methods are shared across request and response definitions, such as `with_header()` and `with_body()`. Pact Python automatically applies them to the correct part of the interaction based on whether they are called before or after `will_respond_with()`. Alternatively, these methods accept an optional `part` argument to explicitly specify whether they apply to the request or response.
+
+#### Running Tests
```
YOU54F, line 112: adding imports at the top, which will make it clear its v2 & base code (v3)

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+The new API entirely replaces both of these approaches with a single, consistent method:
+
+```python title="v3"
+pact = Pact("my-consumer", "my-provider")
+
+with pact.serve() as srv:
+    response = requests.get(f"{srv.url}/users/123")
+```
+
+The server host and port can be specified in `server()` if needed, but by default, the server binds to `localhost` on a random available port. More details can be found in the [API reference][pact.pact.Pact.serve].
```
YOU54F: typo, should this be `serve`?

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+
+##### Writing Pact Files
+
+Since the old v2 API executed a sub-process for the mock service, the Pact file was automatically written when the context manager exited or when `pact.verify()` was called. The new v3 API runs the mock service in-process, so the Pact file must be written explicitly using the `write_file()` method:
```
YOU54F, line 162: why does the user need to call write_file, explicitly? Could they forget

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+    .add_transport(
+        protocol='http',
+        port=8080,
+        path='/api/v1',
+        scheme='https'
+    )
+)
+```
+
+#### Adding Pact Sources
```
YOU54F, line 241: headings don’t render properly here

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+
+Support for both local files and Pact Brokers is retained in v3, with the `verify_pacts` and `verify_with_broker` methods replaced by a more flexible source configuration. This allows multiple sources to be combined, and selectors to be applied.
+
+<!-- markdownlint-disable code-block-style -->
+
+=== "Local Files"
```
YOU54F, line 247: can this be a url , for verifications by webhook?

how do we add auth in that case?

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+    ```
+
+    ```python title="v3"
+    # Option 1: URL-based (similar to v2)
+    verifier = (
+        Verifier('my-provider')
+        .add_transport(url='http://localhost:8080')
+        .state_handler(
+            'http://localhost:8080/_pact/provider_states',
+            body=True
```
YOU54F: it is worth commenting on what the body option does here in the v3 example

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+        )
+        .add_source('./pacts/')
+    )
+    ```
+
+=== "Functional State Handling"
```
YOU54F, line 334: is setup/teardown supported for provider states?

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+
+    ```python title="v2"
+    # Not supported
+    ```
+
+    ```python title="v3 - Function"
+        def handler(name, params=None):
```
YOU54F, line 341: might be nice to add example of provider states with params

MIGRATION.md
```diff
@@ -0,0 +1,413 @@
...
+    )
+    ```
+
+    !!! warning
+
+        These functions run in the test process, so any side effects must be properly shared with the provider. If using mocking libraries, ensure the provider is started in a separate thread of the same process (using `threading.Thread` or similar), rather than a separate process (e.g., using `multiprocessing.Process` or `subprocess.Popen`).
+
+<!-- markdownlint-enable code-block-style -->
+
+#### Message Verification
```
YOU54F, line 379: add v2/v3 imports here to make it more obvious

docs/consumer.md
```diff
@@ -1,356 +1,359 @@
...
 
-The consumer is the client that makes requests, and the provider is the server that responds to those requests. In most straightforward cases, the consumer is a front-end application and the provider is a back-end service; however, a back-end service may also require information from another service, making it a consumer of that service.
+The core logic is implemented in Rust and exposed to Python through the [core Pact FFI](https://github.com/pact-foundation/pact-reference). This will help ensure feature parity between different language implementations and improve performance and reliability. This also brings compatibility with the latest Pact Specification (v4).
+
+> [!NOTE]
+>
+> For asynchronous interactions (e.g., message queues), the consumer refers to the service that processes the messages. This is not covered here, but further information is available in the [Message Pact](https://docs.pact.io/getting_started/how_pact_works#non-http-testing-message-pact) section of the Pact documentation.
```
YOU54F, line 40: it might be nice to add a python message example here. the main docs provided context but don't help a python user get started. maybe even just a link to the examples (just checked, I thought we had a v3 message example consumer and provider)

see 

https://github.com/pact-foundation/pact-js/blob/master/docs/messages.md

and 

https://github.com/pact-foundation/pact-php/blob/master/docs/messages.md

MIGRATION.md
JP-Ellis: I think you were looking at this in the GitHub-rendered docs? These will rendered by MkDocs and have titles:

````markdown
```python title="v2"
with pact:
    # tests...
# Pact file written automatically
```

```python title="v3"
pact = Pact('consumer', 'provider')
# Define interactions and run tests...
pact.write_file('/path/to/pacts')
```
````

Which renders as follows:

image

Let me know if you still think the imports are needed for each block though.

File: MIGRATION.md (hunk 2 above)
Comment (JP-Ellis) [reply MIGRATION.md]: Nice catch! Fixed

File: MIGRATION.md (hunk 3 above)
Comment (JP-Ellis) [reply MIGRATION.md:162]: The location where files are written needs to be explicitly given _at some point_, so I don't see a way of fully automating that. I also personally prefer to have side effects like this be more explicit than implicit.

I guess it could be an argument to `Pact.__init__`, but then there's a question as to when the files should be written? The only option I can see is at the end of each `pact.serve()` contexts, but I'm not sure if that's best.

I'm happy to have you weigh in on this if you feel strongly one way or another.

File: MIGRATION.md (hunk 4 above)
Comment (JP-Ellis) [reply MIGRATION.md:241]: These will be tabs in the properly rendered output 👍

File: MIGRATION.md (hunk 6 above)
Comment (JP-Ellis) [reply MIGRATION.md]: Noted, and added a note about it.

File: MIGRATION.md (hunk 1 above)
Comment (YOU54F) [reply MIGRATION.md:112]: ahh yeah, I was initially, maybe having something at the top of the doc that says viewing this on github? try the pact-python website for a full rendered experience or something to that effect


possibly on the imports, it means i could lift and shift straight from the v3 block to my ide, without having to manually add the imports, but its also fine without for brevity.

File: MIGRATION.md (hunk 7 above)
Comment (JP-Ellis) [reply MIGRATION.md:334]: Added a note to explain this

**Review by YOU54F — COMMENTED** [review]
added comment about msg pact, forgot to submit review as had to go into few meetings

File: MIGRATION.md (hunk 8 above)
Comment (JP-Ellis) [reply MIGRATION.md:341]: I've added that on the consumer side

File: MIGRATION.md (hunk 9 above)
Comment (JP-Ellis) [reply MIGRATION.md:379]: I think that's the same as above, with once rendered properly, it should be better; but let me know if you think the imports are needed in addition to the code block titles.

File: docs/consumer.md (hunk 10 above)
Comment (JP-Ellis) [reply docs/consumer.md:40]: Absolutely this needs to be added. I was focusing on reworking what we already had. I might do this as a separate PR though.

We did have an example of the new message pacts in the examples, I just need to refactor it into a standalone example that can be run on its own.

**Comment by JP-Ellis** [comment]
Thanks @YOU54F for the thorough feedback!

I've incorporated your changes and have updated the migration guide.

As for the message-related docs, I want to address that in a separate PR as that is entirely new for V3.

File: MIGRATION.md [hunk 11]
```diff
@@ -0,0 +1,423 @@
...
+
+Message verification is now much more straightforward in v3:
```
YOU54F: it might be worth noting that this closes a few gaps in the previous impl

some notes

https://github.com/YOU54F/pact-logical-replication/tree/main/message_pact_python#pact-python-messaging

might be worth mentioning in the migration guide, that message handlers are correctly mapped to descriptions, not providerStates, and v3 matching rules will be applied where previous v2 matching rules were.

Or closing off those linked issues as being resolved in V3, and referenced here.

nice way to mop up a few outstanding issues, and probably lead users towards using pp v3 if they want to use message pact (and I have seen it in use at customer sites)

docs/consumer.md
YOU54F: Okay dude, there were message pact docs albeit buried in an example readme in the previous core, people had found it and were using it. If it doesn't follow that far behind, its all good

MIGRATION.md
YOU54F: this is the last clarification I really want to mop up before we box this off. If we don't have this functionality, then we can just ticket it up and work forward. If the func does exist, it would be great to document, as to reach pact nirvana, you have two provider verification flows, one by url (triggered by webhook) and one by querying the pact broker with selectors which returns dynamic results.

This was a distinct gap in V2

https://github.com/pact-foundation/pact-python/issues/332

MIGRATION.md
YOU54F: Pretty much all of the pact frameworks have the user set the directory up front as a option in the pact constructor.

They then wrap up the writing of the pact file, with the mock server verification process (so if there are no verification mismatches, then write the file)

Could the user feasibly call write_pact without checking the verification results?

> The only option I can see is at the end of each pact.serve() contexts, but I'm not sure if that's best

that is what the rust consumer core does, pact-js does similar (it performs the test execution in a execute block, and if the verification is successful, the pact is written.

MIGRATION.md
```diff
@@ -0,0 +1,423 @@
...
+    .given('user exists', id=123, name='Alice')  # (1)
+    .with_request('GET', '/users/123')
+    .with_header('Accept', 'application/json')
+    .with_query_parameter('include', 'profile')
+    .will_respond_with(200)
+    .with_header('Content-Type', 'application/json')
+    .with_body({'id': 123, 'name': 'Alice'}, content_type='application/json'))
+```
+
+1.  The new provider states can be streamlined by parameterizing them directly in the `given()` method. So instead of defining multiple variations of a `"user exists"` state, you can define it once and pass different parameters as needed. These can be passed as keyword arguments to `given()`, or as a dictionary in the second positional argument.
```
mefellows: I didn't quite follow this.

Is

```
.given('user exists', id=123, name='Alice') 
```

equivalent to:
```
.given('user exists', id=123) 
.given('user exists', name='Alice') 
```
?

MIGRATION.md
mefellows: The other languages AFAIK only write the contract if the tests pass, so this element has mostly been moved into an internal detail.

There is the problem, however, that the pact file will be _appended_ to by default (as the core allows parallel tests, it doesn't know when the tests start/finish - at least, it doesn't at the moment). 

So having it explicit isn't an issue. 

The detail of the pact file not being overridden by default is worth calling out (It's possible you do further down, but noting it here for now).

MIGRATION.md
mefellows: Good shout Yousaf, we can close that off with this release too!

MIGRATION.md
```diff
@@ -0,0 +1,423 @@
...
+            password='password'
+        )
+    )
+
+    # Or with selectors for more control
+    broker_builder = (
+        verifier
+        .broker_source(
+            'https://pact-broker.example.com',
+            selector=True
```
mefellows, line 291: What is this selector? Does this enable the use of the fluent APIs afterwards or something? If it's set to false, do those function calls get ignored?

MIGRATION.md
mefellows: This is nice, somebody recently asked for a similar one in Pact JS (functional mapping, rather than keyed by state). One downside with this approach, is that the framework can't automatically detect a missing state handler and log a warning to alert the user to it.

MIGRATION.md
```diff
@@ -0,0 +1,423 @@
...
+        These functions run in the test process, so any side effects must be properly shared with the provider. If using mocking libraries, ensure the provider is started in a separate thread of the same process (using `threading.Thread` or similar), rather than a separate process (e.g., using `multiprocessing.Process` or `subprocess.Popen`).
+
+<!-- markdownlint-enable code-block-style -->
+
+#### Message Verification
+
+Message verification is now much more straightforward in v3:
+
+```python title="v3 - Functional Handler"
+def message_handler(description, metadata):
```
mefellows, line 384: Do message handlers support returning (i.e. verifying) metadata?

Example Pact JS: https://github.com/pact-foundation/pact-js/blob/master/docs/messages.md#provider-producer

docs/consumer.md
```diff
@@ -1,356 +1,359 @@
...
+ The `pact.serve()` context manager starts a mock provider service that handles requests according to the defined interactions, creating a controlled testing environment. The actual client code is then executed against this mock service, ensuring it makes correct requests and handles responses properly. Once the context is exited, the Pact file is automatically written to the specified directory for later provider verification, completing the consumer-driven contract testing cycle.
 
-## Requests
+> [!WARNING]
+>
+> A common mistake is to use a generic HTTP client (e.g., `requests`, `httpx`, etc.) to make requests to the mock service within the test. This defeats the purpose of the test, as it does not verify that the client is making the correct requests and handling the responses correctly.
 
-The expected request in the example above is defined with the `with_request` method. It is possible to customize the request further by specifying the method, path, body, headers, and query with the `method`, `path`, `body`, `headers` and `query` keyword arguments.
+### Multi-Interaction Testing
```
mefellows, line 156: Nice 👌

docs/consumer.md
```diff
@@ -1,356 +1,359 @@
...
+Generators are particularly useful when:
 
-## Broker
+-   **Testing with fresh data**: Each test run uses different values, helping catch issues with data handling
+-   **Avoiding test pollution**: Dynamic IDs and tokens prevent tests from accidentally depending on specific values
+-   **Simulating real conditions**: Generated timestamps, UUIDs, and random numbers better represent actual API behavior
+-   **Provider state integration**: Using `generate.provider_state()` to inject values from the provider's test setup
 
-The above example showed how to test a single consumer; however, without also testing the provider, the test is incomplete. The Pact Broker is a service that allows you to share your contracts between your consumer and provider tests.
+### Common Generators
```
mefellows, line 300: Do the Python generators also allow passing a value for the _consumer_ side? One (big) downside of generators, is that they can generate random values in the consumer tests each time they run, which results in an otherwise static contract having a new hash. This has the effect of cancelling the pre-verification feature of the broker, requiring a provider verification.

The way a user can mitigate this is to pass a static value for the consumer side, whilst allowing the generator to use a dynamic value at provider verification time.

docs/provider.md
```diff
@@ -1,158 +1,294 @@
...
+
+def test_provider_with_selectors():
+    """Test with advanced broker selectors."""
+    verifier = (
+        Verifier("my-provider")
+        .add_transport(url="http://localhost:8080")
+        .broker_source(
+            "https://my-broker.example.com",
+            token="bearer-token",
+            selector=True,  # Enable selector builder
```
mefellows, line 97: I think this answers my question from above.

docs/provider.md
```diff
@@ -1,158 +1,294 @@
...
-    | ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-    | `--pact-url TEXT`            | specify pacts as arguments instead. The URI of the pact to verify. Can be an HTTP URI, a local file or directory path. It can be specified multiple times to verify several pacts. |
-    | `--pact-urls TEXT`           | specify pacts as arguments instead. The URI(s) of the pact to verify. Can be an HTTP URI(s) or local file path(s). Provide multiple URI separated by a comma.                      |
-    | `--provider-states-url TEXT` | URL to fetch the provider states for the given provider API.                                                                                                                       |
+```python
+verifier = (
+    Verifier("my-provider")
+    .add_transport(url="http://localhost:8080")
+    .broker_source("https://my-broker.example.com", token="bearer-token")
```
mefellows, line 119: Does the verifier also respect the standard `PACT_BROKER_BASE_URL` and `PACT_BROKER_TOKEN` environment variables?

docs/provider.md
```diff
@@ -1,158 +1,294 @@
...
+    Verifier("my-provider")
+    .add_transport(url="http://localhost:8080")
+    .add_source("./pacts/")
+    .state_handler(handle_provider_state, teardown=True)
+)
+
+verifier.verify()
+```
+
+### Dictionary-Based State Handler (Recommended)
```
mefellows, line 224: Out of curiosity, why do you see this as the recommended approach?

docs/consumer.md
```diff
@@ -1,356 +1,359 @@
...
 ```
 
-In this example, the `UserClient` must include a `X-Request-ID` header matching the pattern (irrespective of the actual value), and the mock service will respond with the example values.
+It expects the following environment variables to be set:
 
-### Like
+`PACT_BROKER_BASE_URL`
 
-The `Like` matcher asserts that the element's type matches the matcher. If the mock service needs to produce an answer, the example value provided will be returned. Some examples of the `Like` matcher are:
+:   The base URL of the Pact Broker (e.g., `https://test.pactflow.io` if using [Pactflow](https://pactflow.io), or the URL to your self-hosted Pact Broker instance).
```
YOU54F: nit: PactFlow

docs/consumer.md
```diff
@@ -1,356 +1,359 @@
...
-    result = client.get_user(123)
-    assert result["id"] == 123
-    assert result["reference"] == "X1234-456def"
-    assert result["last_modified"] == "2024-07-20T13:27:03Z"
+```console
+pact-broker publish \
+    /path/to/pacts/consumer-provider.json \
+    --consumer-app-version 1.0.0 \
+    --tag-with-git-branch \
+    --auto-detect-version-properties
```
YOU54F, line 221: lets avoid tags if we can, and remove the consumer-app-version, as its preferred there is at least a sha identifier as part of the version, which `auto-detect-version-properties` will pick up

docs/provider.md
```diff
@@ -1,158 +1,294 @@
...
-    | `--pact-urls TEXT`           | specify pacts as arguments instead. The URI(s) of the pact to verify. Can be an HTTP URI(s) or local file path(s). Provide multiple URI separated by a comma.                      |
-    | `--provider-states-url TEXT` | URL to fetch the provider states for the given provider API.                                                                                                                       |
+```python
+verifier = (
+    Verifier("my-provider")
+    .add_transport(url="http://localhost:8080")
+    .broker_source("https://my-broker.example.com", token="bearer-token")
+    .set_publish_options(
+        version="1.2.3",  # Provider version
+        branch="main",
```
YOU54F, line 125: it is worth stating it is recommend the version / branch are taken from git and not static

docs/provider.md
```diff
@@ -1,158 +1,294 @@
...
-    | ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-    | `--pact-url TEXT`            | specify pacts as arguments instead. The URI of the pact to verify. Can be an HTTP URI, a local file or directory path. It can be specified multiple times to verify several pacts. |
-    | `--pact-urls TEXT`           | specify pacts as arguments instead. The URI(s) of the pact to verify. Can be an HTTP URI(s) or local file path(s). Provide multiple URI separated by a comma.                      |
-    | `--provider-states-url TEXT` | URL to fetch the provider states for the given provider API.                                                                                                                       |
+```python
+verifier = (
+    Verifier("my-provider")
+    .add_transport(url="http://localhost:8080")
+    .broker_source("https://my-broker.example.com", token="bearer-token")
+    .set_publish_options(
```
YOU54F: how does one enable conditional publishing? such that they can run this test on their local machine and not publish results, but only publish from CI.

Is there a publishVerificationResult option, which the user can setup to be controlled by checking the env.CI flag?

YOU54F (commented):
last round of comments, nothing too big

docs/consumer.md
YOU54F: Agreed :)

MIGRATION.md
YOU54F: Could this be omitted from the end user interface, and they just can just chain a selector builder, or not.

I assume if the selector option is False, or not provided, we will just get the recommended set of selectors back from the PB (deployed/released/main branch), but I probably want enablePending/includeWip pacts to align with the pact nirvana recommendations, so I would probably always have this selector option enabled.

MIGRATION.md
YOU54F: I think it is saying you could provide any number of named args after the 1st, or a dict as the 2nd, to result in 


```
providerStates:[ {"name":"user exists", "params":{"id": 123, "name":"Alice"}}]
```


If you were doing this

```
.given('user exists', id=123) 
.given('user exists', name='Alice') 
```

I would expect this in the pact file (2 distinct states, as they are considered unique due to the the diff params)

```
providerStates:[ {"name":"user exists", "params":{"id": 123}},{"name":"user exists", "params":{ "name":"Alice"}}]
```

MIGRATION.md
JP-Ellis: I don't call that out in the migration guide, but the `write_file` option does have an `overwrite: bool = False` option.

I think I might leave that as an explicit option then.

MIGRATION.md
JP-Ellis: I have no tested the webhook flow at all... so I'll take a look and see.

MIGRATION.md
JP-Ellis: The Python implementation raises an error if the state isn't known and the test stops.

To know if there's a missing state, the provider would need to list out all states from all consumers, which I think was only recently added in the Pact Broker's API?

docs/consumer.md
JP-Ellis: Yeah, I have the placeholder for the [message example](https://pact-foundation.github.io/pact-python/examples/#message-examples) and the old code [still exists](https://github.com/pact-foundation/pact-python/blob/main/examples/v2/src/message.py), it just needs to be migrated over to the new format. Not a huge deal since the code already is there.

I was thinking of adding a separate "Message Pact" doc that goes over specific of messages, assuming the user has read both consumer and/or provider docs; as there is a lot of overlap already.

MIGRATION.md
JP-Ellis: My god, I forgot about the description/provider state crossed wire 😆

But good point though! I've been focusing so much on V3 that I've forgotten some of the weirdness of V2 😆

MIGRATION.md
JP-Ellis: Yeah, re-reading my wording, it isn't clear, what I mean is that you can have a single state called `"user exists"` and parameterize it, instead of having `"user Alice exists"` and `"user Bob exists"` (which would quickly lead to an unnecessary proliferation of provider states).

And looking into it further, I may in fact be incorrect in the implementation because of a subtle difference between `given_with_param` and `given_with_params`:

- The underlying [`given_with_param`](https://github.com/pact-foundation/pact-reference/blob/fa52876a2866e2740e7ad9173a934df25178d0a1/rust/pact_ffi/src/mock_server/handles.rs#L621) FFI **merges** all of the parameters under the one provider state, so in terms of the FFI (in some pseudo-python)

  ```python
  given_with_param('user exists', id=123)
  given_with_param('user exists', name='Alice')
  ```

  will result in just _one_ provider state with two parameters.

- On the other hand, [`given_with_params`](https://github.com/pact-foundation/pact-reference/blob/fa52876a2866e2740e7ad9173a934df25178d0a1/rust/pact_ffi/src/mock_server/handles.rs#L666) seems to **append** a new state each time, so in terms of the FFI:

  ```python
  given_with_params('user exists', id=123)
  given_with_params('user exists', name='Alice')
  ```

  will result in _two_ separate provider states, both with a single parameter.

I think the best outcome is to use the _latter_ option, so that we can create multiple users at once:

```python
(
	interaction
	.given('user exists', id=123, name='Alice)
	.given('user exists', id=234, name='Bob')
)
```

will ensure that there are _two_ provider states, and both Alice and Bob exist.

MIGRATION.md
JP-Ellis: You are correct, `selector=True` then returns a `BrokerSelectorBuilder` instance where the various attributes can be set. I'll clarify that both in the migration doc, and the API docs

MIGRATION.md
JP-Ellis: Yes, it should be able to support these.

docs/consumer.md
JP-Ellis: Yeah, in most cases, a default value is either highly recommended, or required. I remember you explaining this behaviour in Pact JS 😄

docs/provider.md
JP-Ellis: I wasn't sure whether to that or not. At the moment, no, the end-user would have to use something like:

```
.broker_source(os.getenv("PACT_BROKER_BASE_URL"), token=os.getenv("PACT_BROKER_TOKEN"))
```

I guess I could tweak the logic so that:

```
.broker_source()
# or
.broker_source(selector=True).include_pending(True).build()
```

would then populate automatically from environment variables.

Any strong opinion on this? I think either are fine.

docs/provider.md
JP-Ellis: I like it better because it is more flexible and legible:

1. I like that the keys can be inspected directly for the list of provider states
2. I like that keys can be added (or even removed) at runtime if need be
3. I like that the target functions called clearly handle a single state
4. I like that the target functions can be defined as lambdas within the code

docs/consumer.md
JP-Ellis: Fixed :)

docs/provider.md
JP-Ellis: Yes, absolutely!

docs/provider.md
JP-Ellis: I would expect something like:

```python
verifier = (
    Verifier("my-provider")
    .add_transport(url="https://localhost:8080")
    .broker_source("https://my-broker.example.com", token="bearer-token")
)

if "CI" in os.environ:
    verifier.set_publish_option(...)

verifier.verify()
```

I'll update this example to showcase this.

JP-Ellis:
I've pushed the latest changes:

- A few typos/minor wording changes here and there
- Update the `broker_source` to use env vars
- Clarified some docs on the `selector` option

I do need to check the `given` implementation to make sure that new provider states get created, as opposed to merely merging them.

UPDATE: last commit should clarify this.

MIGRATION.md
mefellows: Yes, that's what I was hoping to clarify here. 

> So instead of defining multiple variations of a `"user exists"` state, you can define it once and pass different parameters as needed.

This could be interpreted as two things:

1. It solves the previous provider state problem that didn't support parameters
2. It automatically expands to multiple provider states, each with different parameters

I think it's (1).

docs/consumer.md
JP-Ellis: So remove the `--tag-with-git-branch` option?

Update: I've removed it from the latest update to this PR.

docs/provider.md
JP-Ellis: I'm implemented populating this from env vars :)

MIGRATION.md
JP-Ellis: This has been clarified in the latest update to the docs.

MIGRATION.md
JP-Ellis: I mean both (1) and (2), so it kinda works that it can be interpreted both ways 😆 👍

MIGRATION.md
JP-Ellis: I've elaborated on that point a bit. I feel like the change from V2 is too large to provide a nice before/after, so I'm still going to omit any V2 examples here.

docs/consumer.md
YOU54F: thank you, I'd like to remove `consumer-app-version` if possible too, or maybe update to be the `<python project version>-<git sha>` if you want to retain it.

docs/consumer.md
YOU54F: yep creating a separate messages.md is probably sensible. will close off this thread and we can do that in a follow up PR

YOU54F (approved):
Looking great now, and we can add a full messages docs in a follow up PR and also look at the webhook flow (pact by url). it isn't something present in v2, so doesn't really need consideration for migration, but definitely helps us be more pact nirvana compliant :)

Great stuff Josh

JP-Ellis:
Merged!

I've created separate tickets to track the outstanding (post v3) docs updates:

- [ ] #1266 
- [ ] #1267

YOU54F:
Nice one.

Just as an FYI, the pact-python doc sync job is failing against docs.pact.io

https://github.com/pact-foundation/docs.pact.io/actions/runs/16982245147/job/48144350883

We may want to update the cover page to reflect some blurb towards pact-python v3 having coverage for all the latest specs using the rust core, and provide a link to the seperate pact python docs site.

maybe that is preferred than replicating the docs pages on the docs.pact.io directly as is currently done for projects

JP-Ellis:
Ah, I hadn't seen that. I've created

- [ ] https://github.com/pact-foundation/docs.pact.io/issues/391

To track that issue.

PR #1316, opened 2025-10-31
chore(docs): api docs link on pact-python site is case sensitive

YOU54F:
https://pact-foundation.github.io/pact-python/API/

and

https://pact-foundation.github.io/pact-python/api/

don't route to the same page, unsure if it should, so update link


JP-Ellis:
Thanks for catching that! I didn't realise it was case sensitive, and I think I changed the convention early one to default to `kebab-case` for the URLs

PR #1584, opened 2026-04-26
compute_wrapper_version crashes on 3-part current_version

fire8cker:
If current_version is a standard 3-part semver like '0.4.28' (no 4th component), accessing current_parts[3] raises IndexError, crashing the release pipeline.

 this fix a small bug.

### DIFF
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -615,15 +615,21 @@ def compute_wrapper_version(upstream_version: str, current_version: str) -> str:
         upstream_version:
             The latest upstream semver string, e.g. `"0.4.28"`.
         current_version:
-            The current 4-part package version, e.g. `"0.4.28.2"`.
+            The current 4-part package version, e.g. `"0.4.28.2"`or `"0.4.28"`.
 
     Returns:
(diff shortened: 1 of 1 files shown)


JP-Ellis:
What problem is this fixing? You mention that:

> If current_version is a standard 3-part semver like '0.4.28' (no 4th component), [...]

But given the versioning scheme used for both the FFI and the CLI wrappers are _always_ using 4 components, that `if` condition is false.


The only exception is the placeholder version `0.0.0` at the very start. See:

- https://pypi.org/project/pact-python-cli/#history
- https://pypi.org/project/pact-python-ffi/#history

But these versions are completely empty, used only to reserve the package and bootstrap the package.

You can also see that the release process is working fine in CI:

- https://github.com/pact-foundation/pact-python/actions/workflows/release-cli.yml
- https://github.com/pact-foundation/pact-python/actions/workflows/release-ffi.yml

JP-Ellis:
Closing this as I don't think it is needed, but feel free to ping me if you think I'm mistaken.

PR #1669, opened 2026-08-03
fix: report message handler failures instead of transport errors

### LINKED ISSUE #1665: Misleading 'error sending request for url' when message_handler dict returns raw payload (missing 'contents' key)
## Summary

When `Verifier.message_handler({description: <dict>})` receives dict values that lack a `"contents"` key (i.e., raw payload dicts instead of `Message` envelopes), the FFI reports every affected interaction as:

```
Request Failed - error sending request for url (http://localhost:<port>/_pact/message)
```

This error reads as a networking / loopback / transport bug — every debugger I know would first check IPv4/IPv6 resolution, port binding, ordering of `add_transport`, etc. **The real cause is a `KeyError: 'contents'` inside `_handler` at `src/pact/verifier.py:434`**, raised inside the `MessageProducer` HTTP server thread. Python's `http.server` catches the exception, prints a traceback to stderr, and drops the TCP connection. The Rust FFI sees the disconnected socket and surfaces "error sending request for url".

In our case (asset-compass PR #241) this cost roughly six hours of investigation — including two false-lead fixes (one merged and reverted) — before someone noticed the stderr traceback under the FFI error.

## Minimal repro

```python
# pip install pact-python==3.4.0 pytest httpx
from pact.verifier import Verifier
import httpx

# Handler returns a raw payload dict — no 'contents' key.
# pact-python 3.x expects a Message envelope: {contents, metadata, content_type}
handlers = {"some-message": {"field": "value"}}  # raw payload, NOT envelope

v = Verifier("test-provider")
v.message_handler(handlers)

# Direct probe: the MessageProducer server IS listening
with v._message_producer:
    url = v._message_producer.url
    try:
        r = httpx.post(url, json={"description": "some-message"}, timeout=5.0)
        print(f"HTTP {r.status_code}")
    except Exception as e:
        print(f"{type(e).__name__}: {e}")
```

**Observed output:**
```
Exception occurred during processing of request from ('127.0.0.1', 53137)
  ...
  File "/.../pact/verifier.py", line 434, in _handler
    contents=val["contents"],
             ~~~^^^^^^^^^^^^
KeyError: 'contents'
----------------------------------------
RemoteProtocolError: Server disconnected without sending a response.
```

The `KeyError` traceback goes to stderr (easy to miss under pytest capture), then the FFI reports the drop as `error sending request for url ...` to whoever is running `Verifier.verify()`.

## Environment

- pact-python 3.4.0 (also reproduced on the 3.4.1 release-branch build, so the queued transport-ordering fix does not address this)
- pact-python-ffi 0.5.4.1
- Python 3.13
- Ubuntu 24.04 (GitHub Actions runner) and macOS 15 (dev machine) — same behavior

## Suggested fix

Anything that surfaces the handler exception distinguishably from a transport failure would be a huge quality-of-life win. Options in decreasing order of thoroughness:

1. **Validate the handler dict shape at registration time.** When `message_handler({...})` is called with a plain dict, iterate entries and require each `dict` value to have a `"contents"` key (or be a callable, or `bytes`). Raise `TypeError` immediately with a clear message pointing at the offending key and the expected envelope shape. This catches the mistake before `verify()` even starts.

2. **Catch exceptions inside `_handler` and return HTTP 500 with a diagnostic body** instead of letting `http.server` drop the connection. The FFI then sees a real HTTP response and can surface the diagnostic. Something like:
   ```python
   try:
       return Message(contents=val["contents"], ...)
   except KeyError as e:
       raise ValueError(
           f"message_handler dict value for {name!r} is missing key {e.args[0]!r}. "
           f"Values must be a Message envelope like "
           f"{{'contents': <bytes>, 'metadata': {{'contentType': ...}}}}."
       ) from e
   ```
   and have the `do_POST` handler catch these and return `500 <error>` rather than raising.

3. **At minimum, document this failure mode in the `message_handler` docstring** — a sentence like *"If a dict value is missing the 'contents' key, verification fails with a misleading 'error sending request for url' from the FFI. This is a known limitation."*

I'd be happy to open a PR for option 1 if that's the direction you'd want.

## Related

- [asset-compass PR #241](https://github.com/Prudentia-Sciences/asset-compass/pull/241) — the fix on our side (wrap payloads in Message envelope + emit both `contentType` and `content_type` metadata keys).
- I checked PR #1637 / `b8ee0eab` — the transport-first FFI fix queued for 3.4.1 does not address this because it only applies when both HTTP and message transports are registered; a message-only setup with a bad handler shape hits the same failure signature regardless.

Endika:

-   [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #1665) and the maintainers have approved on my working plan.
    

## :memo: Summary

`MessageProducerHandler.do_POST` sent the `200 OK` response line *before* calling
the handler:

```python
self.send_response(200, "OK")

message = self.server.handler(description, data)
```

Any exception raised by the handler therefore left the client with a
half-written response. `http.server` printed a traceback to stderr and dropped
the connection, and the Pact core reported the disconnect as
`Request Failed - error sending request for url (http://localhost:<port>/_pact/message)`.

Three unrelated failures collapsed into that one misleading message:

| Cause | Real exception | Reported as |
| --- | --- | --- |
| Handler dict value missing `contents` | `KeyError: 'contents'` | `error sending request for url` |
| Message description with no handler | `KeyError: '<description>'` | `error sending request for url` |
| Any exception in a user-provided handler | e.g. `RuntimeError` | `error sending request for url` |

This PR makes three changes:

1.  **`MessageProducerHandler.do_POST` calls the handler before responding**, and
    reports any exception as a `500` whose body carries the exception type and
    message. This also makes it consistent with `StateCallbackHandler.do_POST`,
    which already called its handler first.

2.  **`StateCallbackHandler.do_POST` catches handler exceptions** the same way.
    Its ordering was already correct, but an exception still dropped the
    connection rather than producing a `500`.

3.  **`Verifier.message_handler` validates dictionary values eagerly.** The case
    in #1665 is a configuration error that is fully knowable when the handler is
    set, so it is now raised as a `TypeError` on the `message_handler(...)` line
    in the user's own test, rather than surfacing much later as a failed
    interaction. The error names the offending message and shows the expected
    envelope shape:

    ```
    TypeError: Message handler for 'some-message' is missing the 'contents' key.
    Dictionary values must be Message envelopes, such as {'contents': b'...',
    'content_type': 'application/json'}, and not the raw payload.
    ```

    A request for a message with no handler now also lists the messages which
    do have one.

`Verifier.message_handler` now raises `TypeError` for a dictionary value that is
neither a callable, `bytes`, nor a `Message` dictionary. Previously such a value
was accepted and only failed during verification.

This only rejects handlers which could never have worked, and `TypeError` was
already the documented behaviour for invalid handler values. Anyone affected was
already getting a failed verification; they now get a clear error earlier

## :fire: Motivation

Closes #1665.

The error the user sees points squarely at the network layer, so the natural
first steps are to check IPv4/IPv6 resolution, port binding, and the ordering of
`add_transport` — none of which are involved. The reporter of #1665 spent
roughly six hours on it, including two false-lead fixes, before spotting the
stderr traceback beneath the FFI error.

## :hammer: Test Plan

Added to `tests/test_server.py`:

-   `test_message_post_handler_raises` — a handler raising `RuntimeError`
    produces a `500` whose body contains the exception message, and the handler
    is called exactly once.
-   `test_callback_post_handler_raises` — the same for the state callback server.

Added to `tests/test_verifier.py`:

-   `test_message_handler_invalid_dict_value` — a dict without `contents` and a
    value of an unsupported type both raise `TypeError` at `message_handler`
    time.
-   `test_message_handler_unknown_message` — the `KeyError` for an unknown
    message names the messages which do have a handler.

Verified locally on Python 3.14 / Linux:

-   `pytest tests/ --ignore=tests/compatibility_suite --ignore=tests/v2` —
    354 passed, 2 skipped.
-   `pytest tests/compatibility_suite` — 167 passed, 12 skipped, including
    `test_v3_message_producer` and `test_v4_message_provider`, which exercise
    the real FFI message-producer path.
-   `ruff check`, `ruff format --check` and `mypy` clean.

I also reproduced the original report end-to-end against a V3 message pact. The
misleading transport error is gone; the verifier now fails the interaction on a
content-type mismatch against the `500` body, and the cause is visible in the
log:

```
Message handler for some-message raised an exception.
...
TypeError: Message handler for 'some-message' is missing the 'contents' key. ...
127.0.0.1 - - [...] code 500, message Message handler failed
```

## :link: Related issues/PRs

Closes #1665.

### Known limitation

The Pact core still matches the `500` response body against the expected
message body, so the verifier output reports a content-type mismatch
(`expected a body of 'application/json' but the actual content type was 'text/html;charset=utf-8'`) rather than the producer error itself. Surfacing the error explanation in the verifier output would require a change in `pact-reference`, so it is out of scope here. Happy to raise that separately if you think it is worth doing — the eager validation in point 3 above means the most common case in #1665 never reaches the FFI at all.

### DIFF
--- a/src/pact/_server.py
+++ b/src/pact/_server.py
@@ -282,9 +282,21 @@ class MessageProducerHandler(SimpleHTTPRequestHandler, Generic[_CM]):
             self.send_error(400, "Bad Request")
             return
 
-        self.send_response(200, "OK")
+        # The handler must be called before the response line is sent. Any
+        # exception it raises has to be reported as a 500, and not leave the
+        # client with a truncated 200 response.
+        try:
+            message = self.server.handler(description, data)
+        except Exception as e:
+            logger.exception("Message handler for %s raised an exception.", description)
+            self.send_error(
+                500,
+                "Message handler failed",
+                f"{type(e).__name__}: {e}",
+            )
+            return
 
-        message = self.server.handler(description, data)
+        self.send_response(200, "OK")
 
         metadata = message.get("metadata") or {}
         if content_type := message.get("content_type"):
@@ -491,7 +503,17 @@ class StateCallbackHandler(SimpleHTTPRequestHandler, Generic[_CN]):
             self.send_error(400, "Bad Request")
             return
 
-        self.server.handler(state, action, params)
+        try:
+            self.server.handler(state, action, params)
+        except Exception as e:
+            logger.exception("State handler for %s raised an exception.", state)
+            self.send_error(
+                500,
+                "State handler failed",
+                f"{type(e).__name__}: {e}",
+            )
+            return
+
         self.send_response(200, "OK")
         self.end_headers()
 
--- a/src/pact/_server.py
+++ b/src/pact/_server.py
@@ -296,21 +296,44 @@ class MessageProducerHandler(SimpleHTTPRequestHandler, Generic[_CM]):
             )
             return
 
-        self.send_response(200, "OK")
+        # The response is serialised in full before the response line is sent,
+        # for the same reason: a malformed message must be reported as a 500,
+        # and not leave the client with a truncated 200 response.
+        try:
+            contents = message.get("contents") or b""
+            if not isinstance(contents, (bytes, bytearray)):
+                msg = (
+                    f"Message handler for {description!r} returned contents of "
+                    f"type {type(contents).__name__}, expected bytes."
+                )
+                raise TypeError(msg)  # noqa: TRY301
 
-        metadata = message.get("metadata") or {}
-        if content_type := message.get("content_type"):
-            self.send_header("Content-Type", content_type)
-            if "contentType" not in metadata:
+            metadata = message.get("metadata") or {}
+            content_type = message.get("content_type")
+            if content_type and "contentType" not in metadata:
                 metadata["contentType"] = content_type
-
-        if metadata:
-            self.send_header(
-                "Pact-Message-Metadata",
-                base64.b64encode(json.dumps(metadata).encode()).decode(),
+            encoded_metadata = (
+                base64.b64encode(json.dumps(metadata).encode()).decode()
+                if metadata
+                else None
             )
+        except Exception as e:
+            logger.exception(
+                "Message from handler for %s could not be serialised.",
+                description,
+            )
+            self.send_error(
+                500,
+                "Message handler failed",
+                f"{type(e).__name__}: {e}",
+            )
+            return
 
-        contents = message.get("contents", b"")
+        self.send_response(200, "OK")
+        if content_type:
+            self.send_header("Content-Type", content_type)
+        if encoded_metadata:
+            self.send_header("Pact-Message-Metadata", encoded_metadata)
         self.send_header("Content-Length", str(len(contents)))
         self.end_headers()
         self.wfile.write(contents)
--- a/src/pact/verifier.py
+++ b/src/pact/verifier.py
@@ -111,6 +111,65 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _missing_contents_msg(name: str) -> str:
+    """
+    Error message for a message handler dictionary without a `contents` key.
+
+    Args:
+        name:
+            The name of the message whose handler is invalid.
+    """
+    return (
+        f"Message handler for {name!r} is missing the 'contents' key. Dictionary "
+        "values must be Message envelopes, such as {'contents': b'...', "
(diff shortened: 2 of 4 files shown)


JP-Ellis (approved):
This looks really good! Thanks for the PR ❤️

JP-Ellis:
btw, I don't believe the CI failures are due to anything you did. I need to look into it, as it is spread across other PRs (e.g., Renovate PRs which don't touch anything in the main codebase).

JP-Ellis (approved):

